PR #29, opened 2020-11-04
Fix examples and build workflow

SuperPaintman:
Hi! I found that build workflow was broken. When `go build -v ./...` is run it doesn't build modules but the main one.

```
go: directory examples/... is outside main module
go: directory tutorial/... is outside main module
```

And a mouse example was broken.


meowgorithm:
Thank you so much!

PR #30, opened 2020-11-04
Add tests for `mouse.go`

SuperPaintman:
Hi! I've added some tests for `mouse.go` and rewrote `parseX10MouseEvent()`.

Maybe you will find them useful. :)

### DIFF
--- a/mouse.go
+++ b/mouse.go
@@ -70,35 +70,21 @@ func parseX10MouseEvent(buf []byte) (m MouseEvent, err error) {
 		bitShift  = 0b0000_0100
 		bitAlt    = 0b0000_1000
 		bitCtrl   = 0b0001_0000
-		bitButton = 0b0010_0000
+		bitMotion = 0b0010_0000
 		bitWheel  = 0b0100_0000
-
-		bitsRelease = 0b0000_0011
 
 		bitsMask = 0b0000_0011
 
-		bitsLeft   = 0b0000_0000
-		bitsMiddle = 0b0000_0001
-		bitsRight  = 0b0000_0010
-		bitsMotion = 0b0000_0011
+		bitsLeft    = 0b0000_0000
+		bitsMiddle  = 0b0000_0001
+		bitsRight   = 0b0000_0010
+		bitsRelease = 0b0000_0011
 
 		bitsWheelUp   = 0b0000_0000
 		bitsWheelDown = 0b0000_0001
 	)
 
-	if e&bitButton != 0 {
-		// Check the low two bits.
-		switch e & bitsMask {
-		case bitsLeft:
-			m.Type = MouseLeft
-		case bitsMiddle:
-			m.Type = MouseMiddle
-		case bitsRight:
-			m.Type = MouseRight
-		case bitsMotion:
-			m.Type = MouseMotion
-		}
-	} else if e&bitWheel != 0 {
+	if e&bitWheel != 0 {
 		// Check the low two bits.
 		switch e & bitsMask {
 		case bitsWheelUp:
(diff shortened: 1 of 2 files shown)


meowgorithm:
Whoa, this is awesome! A lot more readable. Great to have the tests, too.  One thing, though: I'm noticing that in the mouse example (`examples/mouse/main.go`) button clicks are being reported as unknown. Mind fixing that?

mouse.go
```diff
@@ -58,45 +58,62 @@ var mouseEventTypes = map[MouseEventType]string{
...
+		bitWheel  = 0b0100_0000
+
+		bitsRelease = 0b0000_0011
+
+		bitsMask = 0b0000_0011
+
+		bitsLeft   = 0b0000_0000
+		bitsMiddle = 0b0000_0001
+		bitsRight  = 0b0000_0010
+		bitsMotion = 0b0000_0011
```
meowgorithm, line 83: I _think_ the motion bits here are actually `0b0010_0011` (was `35` as an integer in the previous version of the code).

mouse.go
```diff
@@ -58,45 +58,62 @@ var mouseEventTypes = map[MouseEventType]string{
...
+
+		bitsWheelUp   = 0b0000_0000
+		bitsWheelDown = 0b0000_0001
+	)
```
meowgorithm, line 87: I wonder if it makes sense to move these consts to the package level so that they don't need to be reallocated every time we process a mouse event?

mouse.go
SuperPaintman: Yeh, but we've already checked 6'th bit by `e&bitButton`:

```
bitButton | bitsMotion = 0b0010_0000 | 0b0000_0011 = 0b0010_0011

```py
# For example right click (0b0010_0010).
# We check this is the button event.
0b0010_0010 & bitButton = 0b0010_0010 & 0b0010_0000 = 0b0010_0000

# Then mask the low two bits.
0b0010_0010 & bitsMask = 0b0010_0010 & 0b0000_0011 = 0b0000_0010

# Then check this bits.
0b0000_0010 == bitsRight = 0b0000_0010 == 0b0000_0010 = true
```

The same with scroll events. I checked all events in tests.

mouse.go
SuperPaintman: Currently it doesn't really matter where to place these constants. It's a good practice to place small constants close to the code that use them.

> so that they don't need to be reallocated every time we process a mouse event?

In Go it works a little bit different:
1. In Go replaces constants with its values on compile time. No matter where you place them.
2. Constants are untyped (which means they don't have size before you assign it with a variable / parameter / field / etc.).
3. (//just imagining//) Even if we place them on top level and Go would you them by reference they would allocate `size_t` (32/64bits) to store pointer :).

---

There is only one reason (in my opinion) to put them on top level if you need them in another function / package.

SuperPaintman:
> Whoa, this is awesome! A lot more readable. Great to have the tests, too. One thing, though: I'm noticing that in the mouse example (`examples/mouse/main.go`) button clicks are being reported as unknown. Mind fixing that?

You right, forgot about drags. Fixed and added test cases for them.

mouse.go
meowgorithm: Ah, thanks for the explanation; I didn't consts works like that in Go. In that case it, yeah, it makes sense where you have them. I can't imagine the rest of the package would need them.

meowgorithm:
Awesome, this is working great for me now.

mouse.go
meowgorithm: You're right again :). Makes sense.

muesli (approved):
LGTM!

muesli:
Thank you @SuperPaintman!

PR #31, opened 2020-11-04
Fix links in documents

lusingander:
Some links to pkg.go.dev in the documentation seems incorrect, so fixed.


meowgorithm:
Oof. Thank you SO much for this one!

PR #35, opened 2020-11-22
Fixed build tags

ololosha228:
Cause, you know, windows is **only one** non-unix operating system, and js literally can't use any ansi colors at all :)


caarlos0:
I think this build tags are not actually needed, since the `_unix` file suffix implies it already 🤔

PR #36, opened 2020-11-22
Fix: Log prefix in files now actually adding

ololosha228:
Dead simple, but now it works as expected, lol

Also, sorry for multiple PRs, fix simple errors like this is way more easy in github website, instead commiting via git.

If you don't allow it, i'll sending PRs from git command.


meowgorithm:
Thank you!

PR #46, opened 2021-01-17
Add a tea.Sequentially command

caarlos0:
The idea is to produce a single command that sequentially runs the given commands, returning the first non-nil msg.

Useful for things like saving the current state before quitting, or saving something and then refreshing a list...

Extracted from: https://github.com/caarlos0/tasktimer


meowgorithm (approved):
LGTM!

PR #48, opened 2021-02-01
Add space after prefix if one doesn't exist

meowgorithm:
Right now the logs read `prefix2021/02/01…`. This PR adds a space if one doesn't exist at the end of the prefix, so logs will now read `prefix 2021/02/01…`.

### DIFF
--- a/logging.go
+++ b/logging.go
@@ -3,6 +3,7 @@ package tea
 import (
 	"log"
 	"os"
+	"unicode"
 )
 
 // LogToFile sets up default logging to log to a file. This is helpful as we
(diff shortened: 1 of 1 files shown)


logging.go
```diff
@@ -23,6 +24,13 @@ func LogToFile(path string, prefix string) (*os.File, error) {
 		return nil, err
 	}
 	log.SetOutput(f)
+
+	// Add a space after the prefix if one doesn't exist
+	finalPrefixChar := prefix[len(prefix)-1]
+	if len(prefix) > 0 && !unicode.IsSpace(rune(finalPrefixChar)) {
```
muesli, line 30: The `len` check should probably be dealt with before accesing `prefx[len(prefix)-1]`.

meowgorithm: Oh, you’re right. Thanks @muesli.

PR #50, opened 2021-02-06
fix code in single text input example

vilmibm:
Noticed some errors in this example code, this ought to fix them.


vilmibm:
nevermind, i was confused by the double import and then further confused from there ^_^()

meowgorithm:
Well to be fair, the double import was pretty confusing, at least that's gone now. I imagine `goimports` sucked in that second import at some point and I didn't even notice.

Thanks for the heads up, Nate!

Issue #24, opened 2020-10-30
Input is still read after program quits


curio77:
If multiple `tea.Program`s are run in succession, each of them appears to set up a system signal handler (for catching SIGINT etc.), but does not clean it up after the program has quit without being interrupted this way.  If you run n programs in succession that each want to catch `"ctrl+c"`, for the n-th such program, you'll need to press <kbd>Ctrl-C</kbd> (n-1-m) times, where m is the number of previous programs that *were* interrupted. Of course, the interrupt-catching functionality should work independently of what has happened before.

meowgorithm:
Thanks for flagging this. If I'm reading correctly, this is something you've verified, correct?

Internally Bubble Tea listens for SIGWINCH only. ^C is handled via putting the terminal into raw mode and reading stdin.

curio77:
I haven't taken a look at your implementation (yet).  I was just assuming you are catching the things like <kbd>Ctrl-C</kbd> via `signal.Notify()` but now learned you're doing this via raw terminal mode.

Anyway, the effect I've tried to describe is that for every previous `tea.Program` that has run, you need to press <kbd>Ctrl-C</kbd> one additional time in order to finally get the corresponding message in the latest-running program.  To reproduce, run a program, then another one that just waits for a `"ctrl+c"` message and observe that you'll need to press that combo twice to register as expected.  If you were to run two programs before that, you'd have to press it three times.

mritd:
I also encountered a problem in my program, any keystroke takes a second time to respond.

meowgorithm:
Good to know. I'll bump this one up in priority.

Could you clarify what's happening with the keystrokes, exactly?

mritd:
When the TUI program is run twice in a row, the second time it runs, the key is pressed twice to respond;

This is my test code:

https://github.com/mritd/bubbles/blob/master/example/prompt/main.go#L12

When m1 is running, there will be no response to the first key press (the Update method is not called), and normal operation is restored after the second key press.

Note: After the "Please input password:" prompt, I pressed the "a" key twice, but the terminal only received it once.

![2020-12-27 13 13 45](https://user-images.githubusercontent.com/13043245/103164394-bb76cb00-4845-11eb-9c73-9200e3df660c.gif)

meowgorithm:
Thanks for that. Can you also post the source code for your prompt program?

mritd:
[...]

PR #53, opened 2021-02-24
Automatically listen for terminal input when standard input is not a TTY

meowgorithm:
This PR automatically captures input as normal if stdin is not a TTY, so long as the user hasn't already specified a custom input. Prior to this update, if data was piped in the Bubble Tea runtime would not capture keystrokes, mouse input and so on by default as input was coming in through a pipe and thus not a TTY.

An example is included in the examples directory called `pipe`, which demonstrates this new functionality.

Note that prior to this update it was possible to achieve the above via custom input functionality introduced in #39.

Related: #24.


meowgorithm:
Dropping this branch and PR as there's a simpler, better way to do this, particularly on Windows.

Issue #49, opened 2021-02-04
Altscreen enter and exit messages


meowgorithm:
Right now the only way to enter and exit the altscreen (full window mode) is via `Program.EnterAltScreen()` and `Program.ExitAltScreen()`. This means that, without employing hacks, one can only enter and exit when starting and quitting programs. It probably makes sense to have library-level commands, similar to `tea.Quit`, for entering and exiting the altscreen to give Bubble Tea programs the opportunity to jump in and out of the altscreen in-program.

kiyonlin:
This is a good idea. So we should provide 2 Msg `tea.EnterAltScreen` and `tea.ExitAltScreen`. User can trigger them in Update if they want to change the alt screen mode.

I think I can submit a PR if you haven't done this.

meowgorithm:
That would be awesome, @kiyonlin. Please do feel free to submit a PR. I’d probably call them `tea.EnterAltScreenMsg` and `tea.ExitAltScreenMsg` for naming consistency (like `tea.KeyMsg` and `tea.WindowSizeMsg`).

meowgorithm:
Actually, @kiyonlin; there’s a branch I should merge before you do this. I’ll do that in a few minutes and let you know when it’s in.

kiyonlin:
Dw, I will do it tomorrow (9 hours later) 😃

meowgorithm:
@kiyonlin Alright, go for it. There are some small changes in how the alt screen is set in #61 (just merged) but it shouldn't change your implementation.

meowgorithm:
Oh, just one more thought: realizing we'll also need two commands in addition to messages which I'd probably call `tea.EnterAltScreen` and `tea.ExitAltScreen`.

meowgorithm:
This has been merged and was released in v0.13.0.

PR #62, opened 2021-03-08
Add commands for toggling the altscreen

kiyonlin:
Implement #49

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ examples/textinputs/textinputs
 examples/views/views
 tutorials/basics/basics
 tutorials/commands/commands
+.idea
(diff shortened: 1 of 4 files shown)


meowgorithm:
This is awesome, thank you so much! I made an example which exposed a few gotchas. I'll push an update to your branch in a sec.

kiyonlin:
I think you should put 
```go
	if p.altScreenActive {
		return
	}
```
In lock area

meowgorithm:
Yeah, you're absolutely right

kiyonlin:
I tried p.Enter, exitMsg, enterMsg, p.Exit but enterMsg failed(the program stack). Maybe you should take a look.

meowgorithm:
Thanks again, @kiyonlin!

PR #64, opened 2021-03-09
WithOutput option should take io.Writer as argument type.

### LINKED ISSUE #63: Why WithOutput accept *os.File type
We have `WithOutput(output *os.File)`, but Program.output is `io.Writer`, is this on purpose?

kiyonlin:
This can fix #63 but cause a breaking change.


muesli:
I think this is fine, as it should be source compatible. Thank you for your contributions! :heart:

PR #77, opened 2021-04-05
Add lock for updating width and height

kiyonlin:
Ran into race condition [here](https://github.com/gonetx/httpit/runs/2267718515?check_suite_focus=true), and this pr can fix it.


muesli:
Thanks again, @kiyonlin!

kiyonlin:
> Thanks again, @kiyonlin!

You are welcome! Looking forward to the release 😄

PR #89, opened 2021-04-29
Use golang.org/x/term

tklauser:
The `golang.org/x/crypto/ssh/terminal` package is deprecated and merely a
wrapper around `golang.org/x/term`. Use the latter directly and avoid the
dependency on the former altogether.


muesli:
Sorry for this being a bit confusing. We've introduced separate `go.mod`s for the examples and tutorials to avoid transitive dependencies. This probably was a bad idea in hindsight and we're going to clean that up.

tklauser:
> Sorry for this being a bit confusing. We've introduced separate `go.mod`s for the examples and tutorials to avoid transitive dependencies. This probably was a bad idea in hindsight and we're going to clean that up.

No worries, thankfully the failing GitHub actions pointed me to the right place :smile: `tutorials/go.mod` should now be updated as well.

muesli:
Thank you, @tklauser!

PR #90, opened 2021-04-29
Add paginator example

eliostvs:
Hi, I'm couldn't found an paginator example so I created one \o/.

Screenshot


meowgorithm:
This is a perfect example. Thank you!

PR #110, opened 2021-06-11
renderer: only clear new line if it's shorter than the old one

fiws:
avoids clearing lines that will get overwritten by new lines anyway. (ie if the new line has the same length or are longer then the old one)

I've tested this with most examples. but only on my system.

### DIFF
--- a/standard_renderer.go
+++ b/standard_renderer.go
@@ -116,7 +116,11 @@ func (r *standardRenderer) flush() {
 			if (len(newLines) <= len(oldLines)) && (len(newLines) > i && len(oldLines) > i) && (newLines[i] == oldLines[i]) {
 				skipLines[i] = struct{}{}
 			} else if _, exists := r.ignoreLines[i]; !exists {
-				clearLine(out)
+                // But only clear the line if it's shorter than the old one,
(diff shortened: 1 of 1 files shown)


meowgorithm:
Just a note that I’ve rebased `master` in the main repository onto this branch.

meowgorithm:
I believe we'd need to actually measure line widths with [`go-runewidth`](https://github.com/mattn/go-runewidth), since `len` measures the number of bytes and we actually need the number of cells ([here](https://github.com/charmbracelet/bubbletea/pull/110/files#diff-eec1160b5c25625c859865f3ea86e45e76521a2dd877c78a508183b968646f7dR121)).

Here’s an example of what I mean: https://play.golang.org/p/RytYplOCRoW

I'm not sure yet if it's worth the small overhead in looking up widths for each line.

fiws:
Makes sense, i have not thought about the character widths.

we can close this if you think it's not worth it, no worries.

meowgorithm:
Just a follow up that I'm going to close this one up. Went back and forth on it a little but ultimately I don't think it's worth the (admittedly small) overhead and code complexity. Thanks for the PR regardless!

PR #120, opened 2021-08-04
Cancelable reads and goroutine cleanup prior to termination

erikgeiser:
This PR makes sure that all goroutines terminate gracefully before `Start()` returns. The motivation for this is to avoid having goroutines survive and consume input events even after the `tea.Program` terminates as this results in unreliable behavior of subsequent `tea.Program`s.

However, the input read loop cannot be reliably aborted in all cases as it usually hangs in blocking `Read()` calls. For unix systems with `*os.File` inputs this is solved by waiting for `Read()` not to block before actually calling it with the `select` syscall. For other systems the goroutine only terminates after the current `Read()` returns.

Before merging, an issue has to be solved where `ctrl+c` events keep being consumed after termination on Windows (see https://github.com/charmbracelet/bubbletea/issues/24). Also, if I find the time, the `select` syscall can be replaced with `epoll` (Linux) and `kqueue` (macOS) to overcome the limitation of `select` that the file descriptor has to be smaller than 1024.


erikgeiser:
Currently we have `epoll` for Linux, `kqueue` for macOS/*BSD, `select` for other Unixes and `WaitForMultipleObjects` for Windows. The latter sometimes does not work in Windows Terminal for some reason. In this case, it cannot cancel the ongoing blocking read but it will make the input loop terminate at least as soon as the read finished. For other esoteric platforms, it will also terminate after the next keypress.

erikgeiser:
I think the Windows issue is either a bug in Windows Terminal or it is related to events that trigger `WaitForMultipleObjects` but are ignored in `ReadFile` but I tried to set the terminal mode to ignore such events (see https://docs.microsoft.com/de-de/windows/console/setconsolemode).

I think it is possible to have a clean Windows implementation based on [PeekConsoleInput]`https://docs.microsoft.com/is-is/windows/console/peekconsoleinput`). In this case we could peek, check if the program wants to terminate and only read the event otherwise. However, this reads [INPUT_RECORD](https://docs.microsoft.com/is-is/windows/console/input-record-str) structures instead of bytes, so the whole `bubbletea` key handling has to be rewritten but only for Windows. This could be done in future PRs if neccessary.

erikgeiser:
The `kqueue` implementation currently does not work with the `pipe` example. It seems like after the piped in data is read, `kqueue` always returns immediately.

erikgeiser:
I think this PR is kind of ready now. I left a debug print in there that prints "read loop hangs" that triggers if a read cancellation does not work unexpectedly. This is kind of a large set of changes so maybe we should discuss some of these and test it on all of our `bubbletea` applications. I already tested most of the examples in this repository and everything seems to work. If that's your experience too I'll remove the debug print before the PR is merged.

meowgorithm:
Sounds good: we'll start taking a look at this. I really appreciate your help with this one, Erik.

erikgeiser:
Small update on the goroutine lifecycles compared to master.

After a `bubbletea` program terminates:
* on current master: Up to 7 goroutines from `bubbletea` still exist, 3 of which will never terminate
* in this PR: No goroutine from `bubbletea` runs anymore

Liberatys:
@erikgeiser Not really my business 😆 but that is some f***ing clean code 👍🏻 ❤️

erikgeiser:
@Liberatys Thanks you! Code style is very important to me so this means a lot to me.

erikgeiser:
In order to reduce termination latency after a quit signal, I had to let the `Cmd` processing (those that execute the `Cmd` and queue the resulting message) goroutines leak. This is due to the fact that `Cmd` can block for some time (like for example half a second for a blink ticker).

Unfortunately, in the current design a `Cmd` cannot be cancelled. Ideally a signature like `type Cmd func(context.Context) Msg` would help with that. This is maybe something to consider if it is time vor a `v2`.

What we can do in this PR, however, is to make an alternative `Start(...)` function that also returns a channel on which the user can wait after termination until all `Cmd`s are processed if this is necessary. This depends on if users actually care about the `Cmd`s that are processed after the quit signal is processed. If not we can leave it as it is an let the `Cmd` processing goroutines leak.

erikgeiser:
I found out that when I flush the console input initially on Windows, the read cancellation seems to work fine even in Windows Terminal.

Raphexion:
@erikgeiser Great PR :tada: . Solved an issue I was having in my code when I ran two programs back-to-back :heart:

erikgeiser:
@Raphexion Happy to help!

muesli:
Nice work @erikgeiser! Thank you for your contributions!

PR #126, opened 2021-09-04
fix: avoid global state

### LINKED ISSUE #91: Example contains unnecessary global state
The example should work just as well if the initialModel is instantiated in func main, rather than as a global variable. Teach good habits, not bad ones :)

caarlos0:
closes #91


muesli:
LGTM!

PR #127, opened 2021-09-04
fix: remove unnecessary build tags

caarlos0:
closes #35

the filename suffix (`_unix` and `_windows`) already do this, no need to specify in build tags too afaik...


muesli:
I don't think "unix" is a valid keyword here. Only specific OS names like "linux" or "windows" are supported.

caarlos0:
oh, indeed, https://github.com/golang/go/issues/20322 

in my head this was already implemented for some reason...

closing

PR #134, opened 2021-09-22
tea: Add StartM() that returns final model

Raphexion:
Sometimes it is useful to obtain the final model after the
program has finished. Add StartM() (Model, error), so that
it is possible to obtain the final model.

Make sure to keep the original Start() error function, so that
everything is backwards compatible.

The M in StartM was inspired by cobra. Which uses Run and RunE.

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -247,8 +247,8 @@ func NewProgram(model Model, opts ...ProgramOption) *Program {
 	return p
 }
 
-// Start initializes the program.
-func (p *Program) Start() error {
+// StartM initializes the program. Returns the final model.
+func (p *Program) StartM() (Model, error) {
 	p.msgs = make(chan Msg)
 	p.done = make(chan struct{})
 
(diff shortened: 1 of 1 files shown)


meowgorithm:
This probably makes sense, though I almost wonder if `Start()` should return the model itself. It's a breaking change, though it potentially keeps things simpler in the long run.

Regardless, this should happen after #120 is merged.

Raphexion:
@meowgorithm Thank you for looking at my PR and giving feedback.

I agree that in the long run it would be better to change the signature of `Start() error` to `Start() (Model, error)`. I am unsure how many people would get really angry. It is a "simple" change but it might break a lot of code. I am okay with both approaches (breaking or backwards compatible)

Raphexion:
@meowgorithm I've rebased on top of master (post #120).

meowgorithm:
This is now available in Bubble Tea [v0.19.0](https://github.com/charmbracelet/bubbletea/releases/tag/v0.19.0). Thanks again: this is a great feature!

Raphexion:
@meowgorithm Thank you for the mention in v0.19.0. The main reason I could contribute was that you have done a great job to start with. The codebase is very approachable and well written. Thank you for all you hard work and your contributions to open-source :sunny: :sunny: :sunny:

PR #141, opened 2021-10-04
Add program-level quit convenience method

meowgorithm:
This update adds a convenience function for quitting Bubble Tea programs from outside a `Program`. In instead of doing this:

```go
p := NewProgram(...)
p.Send(tea.Quit())
```

You can just…

```go
p := NewProgram(...)
p.Quit()


rohanshiva:
Can you please add an example of this command?

PR #145, opened 2021-10-28
Example calendar component

gfeun:
Linked to the addition of a new calendar bubble here: https://github.com/charmbracelet/bubbles/pull/76

To test see `go run examples/calendar/main.go`


bashbunni:
Here's a screenshot of what this looks like in its current state:
image

I think it's a cool idea! I think we would need to tailor the styling to make it more streamlined with our other bubbles before merging. Awesome work though, thank you for your contribution! I can't speak to when it will be merged, but I will mention it :)

PR #164, opened 2021-12-06
Fix links in command tutorial README.md

garrettgsb:
Noticed some links with incorrect URLs in the README. I updated them to working versions that correspond to what I think they were supposed to point to.


muesli:
Are you sure these links are broken for you? These are footnote references and should be resolved correctly.

muesli:
@garrettgsb Apologies, I just realized these footnote references were indeed missing in the commands tutorial. This should have been fixed by #177 now.

garrettgsb:
Cool, all of the links that I had changed are working now! Will close :)

PR #167, opened 2021-12-09
Add cancelreader bsd go1.17 compilation flags

aymanbagabas:
and fix "in switch on ident (mismatched types uint64 and uint32)" error
when building for 32-bit bsd


muesli:
LGTM!

PR #180, opened 2022-01-04
fix(tea): allocate msgs channel in constructor

paralin:
Race condition bug: Start() is called in a new gorotuine, then Send(). If the Send happens before the msgs channel is allocated, the message is dropped.

Instead allocate the channel in the constructor, so msgs is never nil.


meowgorithm:
Looks good and makes sense. Thanks for the fix!

PR #214, opened 2022-02-01
Handle nil cmds in tea.Sequentially

ajeetdsouza:
There are functions which may or may not return a `tea.Cmd` when they're done. Currently, I have to handle these nil values myself, leading to a lot of extra code.

It would be fantastic if `tea.Sequentially` would handle this for me.


caarlos0:
hey, this is a good catch, thank you!

PR #217, opened 2022-02-03
feat: tea.Batch returns nil if all cmds are nil

caarlos0:
a different approach to  #200


muesli (approved):
LGTM! Appreciate the tests!

PR #218, opened 2022-02-03
feat: allow passing a custom panic handler

aymanbagabas:
Use `WithPanicHandler()` to pass a function that gets executed on panics

### DIFF
--- a/options.go
+++ b/options.go
@@ -130,3 +130,11 @@ func WithANSICompressor() ProgramOption {
 		p.startupOptions |= withANSICompressor
 	}
 }
+
+// WithPanicHandler sets the program's panic handler. This won't have any effect
+// if the WithoutCatchPanics option was enabled.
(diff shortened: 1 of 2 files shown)


meowgorithm:
If you don’t mind, let’s pull `WithPanicHandler` from the PR. Would prefer to consider it a bit more before we expose it. Everything else looks good, but please do test this a bit and let’s hold off on putting it into a release.

aymanbagabas:
> If you don’t mind, let’s pull `WithPanicHandler` from the PR (unless it’s needed right now for Soft Serve). Otherwise, would prefer to consider it a bit more before we expose it. Everything else looks good, but please do test this a bit and let’s hold off on putting it into a release.

Sounds good to me! `Kill()` is now in a separate PR https://github.com/charmbracelet/bubbletea/pull/219

PR #222, opened 2022-02-06
Move cancelreader into a separate package

muesli:
Cleaned up the API and docs a bit and moved CancelReader into a separate package at https://github.com/muesli/cancelreader. This hopefully gives it a bit more exposure and makes it re-usable in other projects (like termenv).


meowgorithm:
Small note there that if we do decide to merge this one it will invalidate #140, so we should consider it carefully.

muesli:
Actually, I think they complement each other quite nicely, together with a slightly adapted #258.

caarlos0 (approved):

Issue #206, opened 2022-01-28
Running programs in subshells


M-Porter:
I just recently switched my project [jumper](https://github.com/M-Porter/jumper) from using tcell to bubbletea but I have run into some issues with ioctl.

Previously i was able to run my script in a sub command like `cd $(jumper to)`. Now when i try that same command with bubbletea i get the the following error:
```
inappropriate ioctl for device
cd: string not in pwd: ^[[?25l^[[?1049h^[[0;0H^[[?1002l^[[?1003l^[[1;m❯^[[0m
```

Some searching in tcell led me to this https://github.com/gdamore/tcell/blob/7946eb8e8d8303b1ee1da909dcdebf448f0b5cb7/nonblock_unix.go#L25-L40

Seems as though bubbletea isn't waiting for the output to drain properly (if im understanding the tcell code correctly.)

meowgorithm:
Good catch and thanks for the report. I can reproduce this on my end. This absolutely something we'll need to fix. We'll keep this thread updated with progress.

M-Porter:
I started digging into it to see if i could figure out how tcell is doing it. Can't promise I can figure out a fix since I don't have much free time anymore but if I can figure out a fix from tinkering around I'll open a PR.

meowgorithm:
@M-Porter thanks for that info. This is definitely something that's not resolved yet and we're very keen to sort this one out. We'll keep this issue updated with progress.

M-Porter:
for sure. i just figured any info along the way would help!

wesen:
I'm very interested in this as well, and have been doing a bit of exploration with muesli's help. I wrote up a few proposals: https://gist.github.com/wesen/045b8ef7227ddd8a6951cd308e5d4634

This seems like a separate issue (that I didn't run into yet), but I might take a closer look, to see if it impacts my use case too.

M-Porter:
@wesen I tested proposal 1 and i can use the program in a subshell now but there are still no colors output.

maaslalani:
Hey @M-Porter, if I'm understanding your issue correctly you want to do something similar to https://github.com/charmbracelet/gum (running Bubble Tea programs and using their output) but running the program separately. They way that we got around this was by running the program on `os.Stderr` and setting the Lip Gloss profile manually `lipgloss.SetProfile(...)` which allows us to do things like:

```
vim $(gum filter)
```

PR #228, opened 2022-02-14
fix: don't check terminal size if we're not writing to a terminal

muesli:
If output isn't a terminal, we can't retrieve the size of terminal. Trying to do so causes an ioctl error.


meowgorithm (approved):

meowgorithm:
Note: this resolves the error in #206, but doesn't solve the issue. After this patch, running Bubble Tea programs in a subshell will produce no output until exit.

PR #237, opened 2022-02-21
add: Exec, program.ReleaseTerminal and RestoreTerminal to re-use input and terminal

### LINKED ISSUE #171: Spawning editors
Hi,

Thanks for the awesome project!

Question: how to properly start subcommand with reuse of Stdout/Stdin?

I'm doing it like so:
https://github.com/antonmedv/llama/blob/58c042eb06a5a661a1388beb4ac09cedf2764ad0/main.go#L183-L190
```go
func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	if m.editMode {
		return m, nil
	}
```
```go
				cmd := exec.Command(lookup([]string{"LLAMA_EDITOR", "EDITOR"}, "less"), filepath.Join(m.path, m.cursorFileName()))
				cmd.Stdin = os.Stdin
				cmd.Stdout = os.Stdout
				// Note: no Stderr as redirect `llama 2> /tmp/path` can be used.
				m.editMode = true
				_ = cmd.Run()
				m.editMode = false
				return m, tea.HideCursor
```
But what to do if a redraw is needed after cmd exits?  And is there a better way to halt the program until cmd is exited? 

Thanks.

_Originally posted by @antonmedv in https://github.com/charmbracelet/bubbletea/discussions/170_

muesli:
`Program.ReleaseTerminal` makes Bubble Tea release the input / terminal, so users can spawn a sub-command. `Program.RestoreTerminal` sets the input reader up again and triggers a repaint.

`Exec` is provided as a `Cmd` interface, and should be the primary method for running a blocking `exec.Command` from within Bubble Tea.

Fixes #171.

### DIFF
--- a/examples/exec/main.go
+++ b/examples/exec/main.go
@@ -49,7 +49,7 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 func (m model) View() string {
 	if m.err != nil {
-		return "Error: " + m.err.Error()
+		return "Error: " + m.err.Error() + "\n"
 	}
 	return "Press 'e' to open your EDITOR.\nPress 'a' to toggle the altscreen\nPress 'q' to quit.\n"
 }
--- a/examples/exec/main.go
+++ b/examples/exec/main.go
@@ -32,7 +32,7 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			return m, cmd
 		case "e":
 			c := exec.Command(os.Getenv("EDITOR")) //nolint:gosec
-			return m, tea.Exec(c, func(err error) tea.Msg {
+			return m, tea.Exec(tea.WrapExecCommand(c), func(err error) tea.Msg {
 				return editorFinishedMsg{err}
 			})
 		case "ctrl+c", "q":
--- a/examples/exec/main.go
+++ b/examples/exec/main.go
@@ -8,9 +8,7 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 )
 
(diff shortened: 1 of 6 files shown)


caarlos0 (approved):

jefferai:
Played around with this for a few minutes this morning and found an issue. I haven't had time to dig into this yet but figured I'd raise it here in case this is going to be merged imminently -- I saw some activity on it yesterday.

I'm running a program that is normally terminated by `Ctrl+C` but accepts both `SIGINT` and `SIGTERM`. If I tell it to terminate by sending it a `kill` then the program terminates and I'm back into my TUI. However, if I terminate it via `Ctrl+C` in the terminal where the program was spawned, I do _not_ go back into my TUI...instead I'm simply left at the CLI.

I followed this only enough to know that it's not hitting the update function's handling of a `ctrl+c` `KeyMsg`.

meowgorithm:
@jefferai Wow, thank you. We were actually planning in merging this. Will look into this, I have some theories as to what’s happening.

caarlos0 (approved):

meowgorithm (approved):

blaggacao:
@muesli 

**Scenario:** 

- As a task runner TUI, I want to run a program (might be interactive or not) and 
- show it's log outputs to the user, 
- as if it was run natrually.

- Upon termintation `0` or `!=0`, I want to give the user the possibilitiy to keep staring at the logs,
- but upon pressing a specified button, e.g. `q` or `esc`
- return to the `tea`TUI

what do you think? Does it make sense?

Issue #179, opened 2021-12-31
Allow unmanaged console output


rfc2119:
Using bubbletea, I would like to have a progress bar similar to `apt`'s:
![apt-progress-bar](https://user-images.githubusercontent.com/4954847/147839573-e081ce74-2908-455e-af17-5ca1af3969ec.png)

Progress bar is updated and the above text does not get cleared. I could not achieve this in bubbletea. The closest attempt I got resulted in an unexpected error (see this [cast](https://asciinema.org/a/V15hrCBa6p9n6U8AHkSde7kkN). the code is a modification of _examples/spinners/main.go_). 

I have also tried using `bubbletea.ScrollDown()` but it instantly clears the screen. Any ideas ?

meowgorithm:
This is a good one definitely a feature we’d like to support, however to do so we’ll need to add some additional functionality to both to the renderer as well as probably a special `Cmd` for unmanaged output.

Anyway, his should be a fun one. We’ll keep this issue updated with progress.

rfc2119:
Thanks for your input @meowgorithm ! If there's anything I can help with, please do not hesitate to ask. I will try to help as much as I can

Adjective-Object:
I have a working approach to this. I tested it out by exposing the `renderer` interface and allowing client programs to replace the renderer with a custom implementation.

Changes to expose the renderer: https://github.com/charmbracelet/bubbletea/pull/246
The renderer: https://github.com/Adjective-Object/tea_log_renderer/blob/main/tea_log_renderer.go

Happy to PR this in @meowgorithm. If you're ok with this approach, let me know if you'd prefer the to keep the Renderer internal and I can make this a feature flag on the standard renderer.

meowgorithm:
Thanks for the info (and the PR!), @Adjective-Object. I'm inclined to keep the rendering interface internal for now as it (somewhat intentionally) hasn't been given the attention necessary for a public API. However, at a glance, your implementation seems to make sense.

Would you be interested in submitting a PR against the standard renderer instead? I've been imagining the public interface as a `Cmd`  or two, perhaps something like:

```go
// A command for printing above the output, with a final newline
func Println(...string) Msg

// A command for printing above the output, probably also with a newline (somewhat like log.Print)
func Printf(string, vars ...string) Msg
```

I'd also _probably_ leave word-wrapping as an exercise for the user, but open to your opinions here.

Let me know what you think!

Adjective-Object:
[...]

PR #246, opened 2022-02-27
Expose the Renderer

Adjective-Object:
Opening this mostly so I can reference it in my comment on #179 

- Exposes the Renderer interface
- Adds the `WithRenderer()` initialization option so users can provide a custom renderer.
- Adds `HandleMessages` to the renderer interface so renderers other than the standard renderer can receive app messages

### DIFF
--- a/nil_renderer.go
+++ b/nil_renderer.go
@@ -2,10 +2,11 @@ package tea
 
 type nilRenderer struct{}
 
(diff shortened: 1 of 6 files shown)


Adjective-Object:
Abandoned in favor of incorporating tea_log_renderer into bubbletea's std renderer. See discussion in #179

PR #249, opened 2022-02-28
Allow unmanaged output above the app in standard renderer

Adjective-Object:
Addresses #179 

- adds Println and Printf to tea, which will log a message above the tea app fold on the next flush
- consolidates consecutive cusrosrUp() ANSI sequences in flush()
 - adds cursorUpBy to support multi-line jumps

### DIFF
--- a/standard_renderer.go
+++ b/standard_renderer.go
@@ -147,14 +147,14 @@ func (r *standardRenderer) flush() {
 				skipLines[i] = struct{}{}
 				jumpedLines++
 			} else {
-				cursorUpBy(out, jumpedLines+1)
-				jumpedLines = 0
-				// otherwise, clear the line so the new rendering can write into it
+				cursorUpBy(out, jumpedLines)
 				clearLine(out)
+
+				jumpedLines = 0
 			}
 		}
 		if jumpedLines >= 1 {
-			cursorUpBy(out, jumpedLines)
+			cursorUpBy(out, jumpedLines+1)
 		}
 
 		if _, exists := r.ignoreLines[0]; !exists {
--- a/standard_renderer.go
+++ b/standard_renderer.go
@@ -138,23 +138,17 @@ func (r *standardRenderer) flush() {
 
 	// Clear any lines we painted in the last render.
 	if r.linesRendered > 0 {
-		jumpedLines := 0
 		for i := r.linesRendered - 1; i > 0; i-- {
 			// If the number of lines we want to render hasn't increased and
 			// new line is the same as the old line we can skip rendering for
 			// this line as a performance optimization.
 			if (len(newLines) <= len(oldLines)) && (len(newLines) > i && len(oldLines) > i) && (newLines[i] == oldLines[i]) {
 				skipLines[i] = struct{}{}
-				jumpedLines++
-			} else {
-				cursorUpBy(out, jumpedLines)
+			} else if _, exists := r.ignoreLines[i]; !exists {
 				clearLine(out)
-
-				jumpedLines = 0
 			}
-		}
-		if jumpedLines >= 1 {
-			cursorUpBy(out, jumpedLines+1)
+
+			cursorUp(out)
 		}
 
 		if _, exists := r.ignoreLines[0]; !exists {
--- a/standard_renderer.go
+++ b/standard_renderer.go
@@ -138,22 +138,23 @@ func (r *standardRenderer) flush() {
 
 	// Clear any lines we painted in the last render.
 	if r.linesRendered > 0 {
-		skippedLines := 0
+		jumpedLines := 0
 		for i := r.linesRendered - 1; i > 0; i-- {
 			// If the number of lines we want to render hasn't increased and
 			// new line is the same as the old line we can skip rendering for
 			// this line as a performance optimization.
 			if (len(newLines) <= len(oldLines)) && (len(newLines) > i && len(oldLines) > i) && (newLines[i] == oldLines[i]) {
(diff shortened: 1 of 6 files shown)


standard_renderer.go
```diff
@@ -192,8 +206,8 @@ func (r *standardRenderer) flush() {
 				_, _ = io.WriteString(out, "\r\n")
 			}
 		}
-		r.linesRendered++
 	}
+	r.linesRendered = numLinesThisFlush
```
Adjective-Object, line 210: newLines might be greater than the size of the appliation, so we can't use it as a loop tracker anymore for the above loop.

Might be worth renaming linesRendered to something like "linesInAppOnLastFlush", or "appHeight" or something similar to avoid confusing the size of the current render with the size we're considering the bubbletea-controlled area of the terminal.

meowgorithm:
Thanks for this! I’m just getting into it; excited to give it a look.

standard_renderer.go
```diff
@@ -122,22 +125,35 @@ func (r *standardRenderer) flush() {
...
 	// Clear any lines we painted in the last render.
 	if r.linesRendered > 0 {
+		skippedLines := 0
 		for i := r.linesRendered - 1; i > 0; i-- {
 			// If the number of lines we want to render hasn't increased and
 			// new line is the same as the old line we can skip rendering for
 			// this line as a performance optimization.
 			if (len(newLines) <= len(oldLines)) && (len(newLines) > i && len(oldLines) > i) && (newLines[i] == oldLines[i]) {
-				skipLines[i] = struct{}{}
-			} else if _, exists := r.ignoreLines[i]; !exists {
```
meowgorithm, line 136: Curious: have we lost the ability to have the renderer ignore lines here? There’s a seldom-used feature here we still need to support where the renderer can be set to skip rendering entirely for certain lines.

meowgorithm:
Good catch. Still testing, but so far it's _really_ good. I'm going to push an example to your branch.

meowgorithm:
Oh also, `Println` and `Printf` will need to be implemented as `Cmd`s (i.e. a `func() Msg`). I can make that edit.

meowgorithm:
It would be easy enough to add some logic to not send messages produced by `Println`/`Printf` commands through the program as well as a not have it trigger a repaint. As a `Cmd` printing would sit well with other “system-level” operations like `Quit` and `EnterAltScreen`.

```go
func Update(msg tea.Msg) (tea.Model, tea.Cmd) {
    // Press any key to send a print
    if _, ok := msg.(tea.KeyMsg); ok {
        return m, tea.Println("Hello!")
    }

    return m, nil
}
```

Adjective-Object:
Alrighty, seems reasonable. My specific scenario only cares about pushing updates from outside the app Update loop, so I suppose it's reasonable that might warrant some additional plumbing.

Is there a way to push a cmd from outside the Update loop? I think I could `program.Send(Println("Hello!")())`, but that strikes me as a bit of a hack.

meowgorithm:
```
```

meowgorithm:
Ah, I see where you're coming from. So for that case like that we may add well add methods on `Program` as well:

```
func (p *Program) Println(string)

func (p *Program) Printf(string, ...string)
```

These would be in-line with the `Program.Quit` helper function. Pushing my small changes and an example in a moment.

Adjective-Object:
This might be a windows termnial only bug, but I noticed an off-by-one error where the skipLines logic was checking the wrong the content at line n-1 when the cursor was on line n.

I think this is because the cursor was skipping back by 1 at the beginning of the printout.

Kind of wish there were a better way to protect this with tests! At any rate, I pushed the fix.

meowgorithm:
Thanks! Yeah, this is definitely a hard one to test. Because it affects something as core as the renderer we'll need to manually check it pretty thoroughly. In my experience, at least, most terminals tend to behave similarly, with the main outlier being the macOS stock terminal which is missing support for a few ANSI sequences.

Adjective-Object:
Looks like the fix caused an off by one in the other direction.

The sequence compresion code is the root of the problem so I've reverted it for now, along with cursorUpBy. Maybe I'll take another shot at it in a separate change.

Adjective-Object:
Going to keep building my project against this branch and see if I can surface any issues. So far it's been stable since 1efd40f0ad70d2a615222fd5e824620f2c151797, I'll let you know if I hit more issues.

meowgorithm:
Sounds good. This will definitely need some good testing before we can merge it.

One question (haven't been able to ascertain it from reading the code yet) is the support for having the renderer completely ignore lines still present?

Adjective-Object:
Skip lines should still work in the latest version, I'm pretty sure.

meowgorithm:
Hey there, @Adjective-Object. Just wanted to check and see if you ran into any more issues with this one. We'll plan on testing a bit more as well: this will be a good one to get in before it gets too stale.

Adjective-Object:
Nope, I've been using this for a few weeks on an internal project and haven't run into any major issues.

meowgorithm:
Okay, awesome. We'll look into this one a bit more then. We need to be fairly thorough we're touching the renderer; that's really the only concern.

maaslalani:
This PR is looking great! I just tested all (29) of the examples in `examples/` (on MacOS + Kitty inside `tmux`) with the new changes and everything works great so I do believe that it doesn't cause anything to break.

And the `package-manager` example works really well :)

meowgorithm:
Just rebased on `master` 〜 this one seems good to go.

PR #250, opened 2022-02-28
fix: minor `tea.Cmd` tutorial syntax typo

wolves:
`errMsg(err)` -> `errMsg{err}`

Just noticed this while reading over the tutorial docs and figured I'd throw up a quick PR to address it.

Really enjoying the tooling so far, thank you for creating it!


meowgorithm:
Good catch. Thanks for the correction and the kind words!

PR #258, opened 2022-03-15
Allow users to supply a custom cancelReader

kylecarbs:
This enables callers to supply their own form of TTY cancellation.

We have our own Windows TTY library based on ConPTY, and would like
to use it with bubbletea, but it's not possible unless we can provide
our own form of read cancellation.

I'm not certain this is the best approach by any means, so suggestions are welcome!

### DIFF
--- a/cancelreader.go
+++ b/cancelreader.go
@@ -30,6 +30,9 @@ type fallbackCancelReader struct {
 // actually cancel an ongoing read but will immediately return on future reads
 // if it has been cancelled.
 func newFallbackCancelReader(reader io.Reader) (cancelReader, error) {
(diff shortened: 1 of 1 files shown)


meowgorithm:
So once we merge #222 we should update this PR to get it where it needs to be in order to merge. Off the cuff, I imagine this would be best implemented as a `ProgramOption`, so something alone the lines of `WithCancelReader(io.Reader)`.

ramezanius (approved):

aymanbagabas:
Hey guys, it's been a bit. Do you still have a need for something like this? If so, what's your use case?

PR #261, opened 2022-03-18
Added Ctrl+(up/down/right/left) + Shift+(up/down/right/left)

Bwahharharrr:
Kept the spacing format used for presentation reasons.
Added shift and ctrl + arrow key identifiers
Was unsure on the keytype numeric values so just took a guess.

### DIFF
--- a/key.go
+++ b/key.go
@@ -137,7 +137,6 @@ const (
 	keyGS  KeyType = 29  // group separator
 	keyRS  KeyType = 30  // record separator
 	keyUS  KeyType = 31  // unit separator
-	keySP  KeyType = 32  // space
 	keyDEL KeyType = 127 // delete. on most systems this is mapped to backspace, I hear
 )
 
@@ -148,7 +147,6 @@ const (
 	KeyEnter     KeyType = keyCR
 	KeyBackspace KeyType = keyDEL
 	KeyTab       KeyType = keyHT
-	KeySpace     KeyType = keySP
 	KeyEsc       KeyType = keyESC
 	KeyEscape    KeyType = keyESC
 
@@ -200,6 +198,7 @@ const (
 	KeyPgUp
 	KeyPgDown
 	KeyDelete
+	KeySpace
 	KeyF1
 	KeyF2
 	KeyF3
@@ -222,8 +221,9 @@ const (
 	KeyF20
 )
 
-// Mapping for control keys to friendly consts.
+// Mappings for control keys and other special keys to friendly consts.
 var keyNames = map[KeyType]string{
+	// Control keys.
 	keyNUL: "ctrl+@", // also ctrl+`
 	keySOH: "ctrl+a",
 	keySTX: "ctrl+b",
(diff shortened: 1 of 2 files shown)


meowgorithm:
Thanks for this! Need to test a bit more before merging but so far it looks good.

key.go
```diff
@@ -139,6 +139,14 @@ const (
...
 	keySP  KeyType = 32  // space
 	keyDEL KeyType = 127 // delete. on most systems this is mapped to backspace, I hear
+	KeyCU  KeyType = 200 // ctrl+up
+	KeyCD  KeyType = 201 // ctrl+down
+	KeyCR  KeyType = 202 // ctrl+right
+	KeyCL  KeyType = 203 // ctrl+left
+	KeySU  KeyType = 204 // shift+up
+	KeySD  KeyType = 205 // shift+down
+	KeySR  KeyType = 206 // shift+right
+	KeySL  KeyType = 207 // shift+left
```
meowgorithm, line 149: You can actually drop these altogether. This section is actually for very specific, predefined ANSI control key sequences.

key.go
```diff
@@ -185,6 +193,14 @@ const (
...
 	KeyCtrlUnderscore   KeyType = keyUS  // ctrl+_
 	KeyCtrlQuestionMark KeyType = keyDEL // ctrl+?
+	KeyCtrlUp           KeyType = KeyCU  // ctrl+up
+	KeyCtrlDown         KeyType = KeyCD  // ctrl+down
+	KeyCtrlRight        KeyType = KeyCR  // ctrl+right
+	KeyCtrlLeft         KeyType = KeyCL  // ctrl+left
+	KeyShiftUp          KeyType = KeySU  // shift+up
+	KeyShiftDown        KeyType = KeySD  // shift+down
+	KeyShiftRight       KeyType = KeySR  // shift+right
+	KeyShiftLeft        KeyType = KeySL  // shift+left
```
meowgorithm, line 203: These can be moved to the "other keys" block below and you can just have the values auto-set via the iota.

Bwahharharrr: If I do this, then for me, "msg.String()" or just `fmt.Println(msg)` is just totally empty.
So key.NewBinding ... WithKeys = .. nothing, because its empty, nothing t o match against.. and switching just msg.String also impossible as no string.

Bwahharharrr: Done

key.go
Bwahharharrr: Done

meowgorithm:
@Bwahharharrr Thanks for the updates! We just merged some other key stuff, so we'll rebase, sort out the conflicts, and then test accordingly.

Bwahharharrr:
After some fighting (obviously I am a git newb), I believe I have rebased with current master  then merged with the new code.

meowgorithm:
Hey there. Just a heads up on where we are with this PR:

* Per #263 we're now erring on using sequences (rather than hex values) for key matching. I'm happy to make the adjustment if you'd like.
* We still need to test against `urxvt`, which often sends different values versus most other terminal emulators and will very likely need it's own set of values to match shift + arrow keys
* We're debating whether or not it makes sense to add a `shift` value to the `Key` struct rather than define things like both `KeyUp` and `ShiftKeyUp`. The `Stringer` interface would still produce values like `"shift+up"`.

meowgorithm:
Hey there. Just a heads up on where we are with this PR:

* Per #263 we're now erring on using sequences (rather than hex values) for key matching. I'm happy to make the adjustment if you'd like.
* We still need to test against `urxvt`, which often sends different values versus most other terminal emulators and will very likely need it's own set of values to match shift + arrow keys
* We're debating whether or not it makes sense to add a `shift` value to the `Key` struct rather than define things like both `KeyUp` and `ShiftKeyUp`. The `Stringer` interface would still produce values like `"shift+up"`.

meowgorithm:
Hey there, @Bwahharharrr. I'm not sure what happened, but I attempted to rebase `master` onto your branch…and when I pushed the issue closed itself. Will reopen in a new PR.

meowgorithm:
Okay, re-opened as #292.

PR #263, opened 2022-03-21
feat: function key support

meowgorithm:
This update adds support for the function keys, such as <kbd>F12</kbd>. Thanks to @mrusme for the request.


mrusme:
Thank you Sir!

Issue #267, opened 2022-03-25
accessing Stdout


tgirod:
For an application that is not using AltScreen, is it safe to have a `tea.Cmd` that writes stuff to Stdout, or could it break the display somehow with some unfortunate concurrent access?

tgirod:
Trying to answer this myself with pull request #268

meowgorithm:
Hi there! This is definitely a feature we'd like to see. There are a number of subtle nuances to getting this right, particularly working in-step with the renderer. That said, there's an open PR for this (#249) which I believe is close, if not there.

Want to give that a try and chime in as to whether or not it's working for you?

meowgorithm:
(Also closing as this is a dup of #179.)

PR #268, opened 2022-03-26
adds Program.Write method

tgirod:
This method gives access to Program.output so one can write directly to the terminal's output.

# Use case

Your app is not fullscreen, and you want to print out some stuff to the terminal while the application is running.

You can do so by using `fmt.Println`, but looking at bubbleatea's code, there is a mutex for accessing `Program.output`. So this method politely grabs the mutex before trying to write to `Program.output`.

# Considerations

Maybe we don't want `tea.Program` to implement the `Writer` interface - in which case the method can be renamed to something else.

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -111,6 +111,14 @@ type Program struct {
 	windowsStdin *os.File //nolint:golint,structcheck,unused
 }
 
+// Write gives direct access to the underlying Writer. Useful for an inline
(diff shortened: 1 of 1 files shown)


meowgorithm:
Closing in favor of #179, as mentioned in #267.

meowgorithm:
Just a note that this functionality is now available in `master` and will be available in the next release. For details see #249.

PR #289, opened 2022-04-11
fix(keys): spacebar now sends a `KeySpace`

bashbunni:
This fixes a bug where the spacebar would send a `Key` with type `KeyRunes` instead of `KeySpace`. The `Runes` portion of `Key` will still contain a space after this fix.

```go
// This is what is sent when you press spacebar now
type Key struct {
    Type: KeySpace,
    Runes: []rune{' '},
}
```

* * *

## Changes:

- [x] remove keySP
- [x] assign KeySpace to " "
- [x] update tests to expect " " instead of "space"

I didn't make any changes to `readInput`. 
Manually tested `KeySpace` in addition to the tests with no issues and `KeySpace` worked as expected

### DIFF
--- a/key.go
+++ b/key.go
@@ -221,8 +221,9 @@ const (
 	KeyF20
 )
 
-// Mapping for control keys to friendly consts.
+// Mappings for control keys and other special keys to friendly consts.
 var keyNames = map[KeyType]string{
+	// Control keys.
 	keyNUL: "ctrl+@", // also ctrl+`
 	keySOH: "ctrl+a",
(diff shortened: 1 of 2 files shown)


meowgorithm:
Thank you, @bashbunni!

Sov3rain:
Hey, I encountered this issue today. I'm on Windows 10, go 1.25.1, Bubbletea 1.3.10. My keyboard is set to French (AZERTY). I ended up checking for both keys like this for it to work:

`case tea.KeyRunes, tea.KeySpace:`

PR #320, opened 2022-05-23
docs: improve godoc on tick and every

caarlos0:
This is always a source of confusion. Thought it would be worth to make it as clearer as possible, not sure if I got it to that point, happy to have some feedback :)


bashbunni:
That looks great! I think that'll help to clarify

meowgorithm:
This is really apt. @caarlos0, I reworded stuff a bit and added an example: lmk what you think.

caarlos0:
looks amazing, thanks @meowgorithm

PR #323, opened 2022-06-01
chore(exec): small API edits

meowgorithm:
Small ergonomic edits to the long-awaited exec API. After this we should be good to do a release.

Side note: I did most of this over the weekend whilst at the dog groomer’s office.


caarlos0 (approved):
looks neat!

PR #325, opened 2022-06-02
docs: default to vim if no EDITOR on exec example

caarlos0:
avoids `Error: fork/exec : no such file or directory` by using vim by default if no EDITOR is set.

### DIFF
--- a/examples/exec/main.go
+++ b/examples/exec/main.go
@@ -11,7 +11,11 @@ import (
 type editorFinishedMsg struct{ err error }
 
 func openEditor() tea.Cmd {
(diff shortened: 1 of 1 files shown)


examples/exec/main.go
```diff
@@ -11,7 +11,11 @@ import (
 type editorFinishedMsg struct{ err error }
 
 func openEditor() tea.Cmd {
-	c := exec.Command(os.Getenv("EDITOR")) //nolint:gosec
+	editor := os.Getenv("EDITOR")
+	if editor == "" {
+		editor = "vim"
```
muesli, line 16: Favoritism! :laughing:

meowgorithm:
Thank you!

PR #329, opened 2022-06-04
feat: allow specifying window size file descriptor

georgijd:
Allow setting the file descriptor to be used when determining the window
size. The default remains the program's stdout. In most cases you won't
need to use this. It's useful in situations where the program's output
file descriptor is not a terminal but the program's stdin might be.
Then it's useful to use the stdin instead as a source of information for
the window size. If the stdin is not a terminal either there isn't much
that can be done and WindowSizeMsg events won't be sent.

### DIFF
--- a/options.go
+++ b/options.go
@@ -130,3 +130,16 @@ func WithANSICompressor() ProgramOption {
 		p.startupOptions |= withANSICompressor
 	}
 }
+
+// WithWindowSizeSource sets the file descriptor to be used when determining
+// the window size. By the default, it's the program's stdout.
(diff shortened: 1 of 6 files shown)


muesli:
We're working on allowing users to specify custom input & output handles, which may superseed this change. You're making valid points however, and I wonder if there's any disadvantage to making the window size detection operate on stdin per default?

georgijd:
Hi @muesli ,

Yes I was thinking of doing an automatic fallback to stdin but decided that it might lead to some unexpected behaviour for some users - window size would start changing when they don't expect it to - so I decided to make it opt-in.

PR #336, opened 2022-06-05
fix: race condition on repaint with alt screen

georgijd:
I didn't realise at the time, but the tea.Program and the renderer share
the mutex. This make the code difficult to reason about - it turns out
the program sometimes acquires the lock and then calls the
`setAltScreen` method of the renderer which in turns calls `repaint`.
That causes a deadlock as `repaint` is trying to acquire the lock too.


muesli:
Yeah, sharing that mutex feels awkward. I think we'll clean that up.

PR #338, opened 2022-06-07
Example: Credit Card Input Form

maaslalani:
Adds an example of a credit card form to `examples`.

```
Total: $21.50:

Card Number
4505 **** **** 1234

EXP     CVV
MM/YY   XXX

Continue ->
```


LONCHHAIHONG (approved):
Visacard ABA Bankcard
Name LON CHAIHONG

PR #343, opened 2022-06-17
support multi-byte strings on Windows

mattn:
On Windows, it use double-byte character sets not UTF-8. So ReadFile return DBCS bytes. it should be converted to UTF-8.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -19,38 +19,38 @@ import (
 // To produce the command, pass a duration and a function which returns
 // a message containing the time at which the tick occurred.
 //
-//	type TickMsg time.Time
+//   type TickMsg time.Time
 //
-//	cmd := Every(time.Second, func(t time.Time) Msg {
-//	   return TickMsg(t)
-//	})
+//   cmd := Every(time.Second, func(t time.Time) Msg {
+//      return TickMsg(t)
+//   })
 //
 // Beginners' note: Every sends a single message and won't automatically
 // dispatch messages at an interval. To do that, you'll want to return another
 // Every command after receiving your tick message. For example:
 //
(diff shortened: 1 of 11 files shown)


muesli:
Hey @mattn!

Nice seeing you over here! Your contribution looks great, but I'm a bit puzzled about the `go fmt` changes. I can't reproduce that output locally, neither with `go fmt` nor with `goimports` or `goreturns`. What tool did you use to format the sources?

If you don't mind, I could revert the formatting changes, so we can focus on reviewing the actual change.

Thanks again!

mattn:
Ah, it seems that specification of `go fmt` changes recently. Sorry. I'll update this PR soon.

mattn:
Hmm, at least, same result in go1.18.

mattn:
I'll make this changes again.

mattn:
https://github.com/golang/go/issues/51082

gofmt is changed

muesli:
> [golang/go#51082](https://github.com/golang/go/issues/51082)
> 
> gofmt is changed

Understood! This seems to be a pending change for 1.19, tho? Can't reproduce it with 1.18.3 for what it's worth. I think we should stick to the existing format until this gofmt change is part of a stable Go release.

mattn:
Reverted format

mattn:
I can squash this changes. Do you want?

muesli:
> I can squash this changes. Do you want?

Thanks, but no need to, can do that here just fine!

meowgorithm:
Small note that this will close #393.

meowgorithm:
Thank you so much for the contribution, @mattn. Multi-byte support is really important to us. Cheers 🍻.

PR #345, opened 2022-06-17
docs: add FAQ

bashbunni:
## Changes:
- [x] add answers to frequently asked questions
- [x] add docs folder


bashbunni:
I no longer think that we need a docs folder. I'd rather include this info either in the README or existing tutorials. Thoughts?

PR #346, opened 2022-06-17
docs: creating nested models

bashbunni:
## Changes:
- [x] add info on working with nested models
- [x] add docs folder

### DIFF
--- a/docs/creating-nested-models.md
+++ b/docs/creating-nested-models.md
@@ -1,0 +1,51 @@
+# Creating Nested Models
+
+There may be situations where you want to have your own nested models for your project. 
+This can allow you to toggle between different views and organize your logic for `Update`.
+
+## Showing Specific Nested Models
+In Bubble Tea programs, you can decide which components are shown by holding a `state` field in your main model struct. 
+For example:
+
+```go
+// this is an enum for Go
+var sessionState uint
+const (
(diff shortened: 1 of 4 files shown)


bashbunni:
hey @maaslalani great suggestion. I actually found some typos while moving the code over to an example :joy: what do you think about these changes? I feel like these docs will become more substantial as we get more questions and as we get feedback. Just pushed the latest changes

maaslalani:
Looks awesome @bashbunni !!!

maaslalani (approved):

meowgorithm (changes requested):
A few quick notes here:

* Please do make sure the example code runs (I'm sure it's still a work-in-progress)
* I would be great to show how one can use a state member in `Update`, in addition to `View`. For example, to change key handling when a section of the app changes focus, since that will be a super common need.

bashbunni:
tbd in dev meeting. I'd like some feedback on writing effective comments in the code and whether or not this example should use `tea.Batch` or return commands. I feel like returning commands makes the flow a bit more obvious, but using `tea.Batch` is the cleaner solution. I can make this PR a single commit as well if that's preferred.

bashbunni:
This example no longer manages state from a main model. That is shown in the `composable-view` branch.

PR #347, opened 2022-06-17
docs: add info on layouts

bashbunni:
## Changes:
- [x] add info on working with layouts
- [x] add docs folder


bashbunni:
I'm creating an example for this instead. I don't think we need a docs directory, we can just add context in different example code or a README.md in `examples/myexample/` if we really need the extra write up

PR #348, opened 2022-06-17
docs: add debounce example

bashbunni:
## Changes:
- [x] add debounce example and context provided from Slack (https://charm.sh/slack, btw)

### DIFF
--- a/examples/debounce/main.go
+++ b/examples/debounce/main.go
@@ -1,0 +1,53 @@
+package main
+
+/*
+With each keypress you bump the tag on the `model` and send that tag along in a `Msg` after delaying with a `tea.Tick`.
+If the tag in the `Msg` matches the one on the model you know the debouncing is done and you can then return your command
+
+We do this in the [spinner](https://github.com/charmbracelet/bubbles/blob/master/spinner/spinner.go) Bubble to keep it from spinning out of control in the event that it accidentally gets too many spin Cmds.
+*/
+
+import (
+	"fmt"
+	"os"
+	"time"
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+type exitMsg int
+
+type model struct {
+	tag int
+}
+
+func (m model) Init() tea.Cmd {
+	return nil
+}
+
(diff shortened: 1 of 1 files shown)


maaslalani:
Let me know if there is a specific reason this is better as markdown though that I haven't considered but I think this should be a `examples/debounce/main.go` so that you can run / modify the example and so that if we ever need to update it we can ensure it still compiles. The "How it works" section is awesome and we can put that at the top of the go file as a comment.

bashbunni:
I agree! Can move it to a running example

bashbunni:
(moved to an example instead of docs)

bashbunni:
let me know if comments could be improved or anything!

examples/debounce/main.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (m model) View() string {
+	return "To exit press any key, then wait for one second without pressing anything."
+}
+
+func main() {
+	if err := tea.NewProgram(model{}).Start(); err != nil {
+		fmt.Println("uh oh:", err)
+		os.Exit(1)
+	}
```
maaslalani, line 52:
```suggestion
    m := model{}
    p := tea.NewProgram(m)
	if err := p.Start(); err != nil {
		fmt.Println("uh oh:", err)
		os.Exit(1)
	}
```

examples/debounce/main.go
```diff
@@ -0,0 +1,53 @@
...
+		if int(msg) == m.tag {
+			return m, tea.Quit
+		}
+	}
+
+	return m, nil
+}
+
+func (m model) View() string {
+	return "To exit press any key, then wait for one second without pressing anything."
```
maaslalani, line 45: Should the view maybe show the `tag` or something so that if someone actually runs this example they know what's going on?

bashbunni: Great idea!

bashbunni:
heh, forgot to rebase from master before pushing 😂

meowgorithm (approved):

meowgorithm:
Thank you @bashbunni!

PR #352, opened 2022-06-21
example:  using the x/exp/teatest package

caarlos0:
Introducing the `teatest` package - a very small package with testing helpers for bubbletea apps.

You basically test a model, giving the model itself, a set of interactions (`tea.Msg`'s) and an assertion, which is basically asserting the output against a given golden file.

Its a very raw idea, any feedback welcome.


purpleclay:
Is this PR a candidate for merging anytime soon? I have been using bubbletea for a little while now, it is a fantastic library, but it does mean you end up with a program that is largely untested through automation. I have been thinking of writing my own library for testing bubbletea, but noticed this PR.

caarlos0:
@purpleclay we're trying it out in some places to see how it feels like... if you wanna give it a try and let us know what you think, it would be greatly appreciated as well (and you can help driving how it should look like).

You can play with it by getting this branch instead:

```sh
go get github.com/charmbracelet/bubbletea@test
```

caarlos0:
@purpleclay you can run `go test ./yourpackage/... -update` and it will generate them for you

purpleclay:
Hi @caarlos0 I have attempted to use `teatest` by just writing a single test case to start with. You can view my test here:

https://github.com/purpleclay/dns53/blob/teatest/internal/tui/dashboard_test.go

So here are my observations (btw I am liking the fact I can test my TUI):

1. My `Update()` method is driven by the initial `Init()` and needs it to return a `tea.Msg` containing EC2 metadata. This requires me to add an enforced sleep. Is this something most people will have to do? As I noticed it in your `simple` example that you do this also. Is there not a way for `teatest` to wait for all `Init` commands to complete automatically?
2. The generation of golden files through the `-update` flag is a nice touch. Would it feel more intuitive to do this through a `go:generate`?
3. Do you envisage people testing partial parts of the terminal output within `teatest.TestModel(func(out []byte) {})`? I imagine I would always use the golden file approach. Attempting to assert against fragments of text would require me to include special characters e.g. `   [1;;mPHZ:[0m             AAAAAAAAAAAAAAAAAAAAAAAA [testing]  `. Partial matching would be useful if these characters could be stripped from the output
4. It seems you cannot run any of the `teatest` test cases if you provide the `-race` flag to `go test`. Is that by design? This is the current output from my test:
```sh
==================
WARNING: DATA RACE
Write at 0x00c0003b7eb0 by goroutine 8:
  bytes.(*Buffer).Write()
      /usr/local/Cellar/go/1.18.3/libexec/src/bytes/buffer.go:169 +0x44
  fmt.Fprintf()
      /usr/local/Cellar/go/1.18.3/libexec/src/fmt/print.go:205 +0xb1
  github.com/charmbracelet/bubbletea.clearLine()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/screen.go:19 +0x85
  github.com/charmbracelet/bubbletea.(*standardRenderer).stop()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/standard_renderer.go:75 +0x31
  github.com/charmbracelet/bubbletea.(*Program).shutdown()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tea.go:588 +0xd1
  github.com/charmbracelet/bubbletea.(*Program).StartReturningModel()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tea.go:496 +0x1cd1
  github.com/charmbracelet/bubbletea.(*Program).Start()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tea.go:549 +0x52
  github.com/charmbracelet/bubbletea/teatest.TestModel.func1()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/teatest/teatest.go:35 +0x53

Previous write at 0x00c0003b7eb0 by goroutine 7:
  bytes.(*Buffer).Write()
      /usr/local/Cellar/go/1.18.3/libexec/src/bytes/buffer.go:169 +0x44
  fmt.Fprintf()
      /usr/local/Cellar/go/1.18.3/libexec/src/fmt/print.go:205 +0xb1
  github.com/charmbracelet/bubbletea.showCursor()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/screen.go:15 +0x4d
  github.com/charmbracelet/bubbletea.Program.restoreTerminalState()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tty.go:30 +0x25
  github.com/charmbracelet/bubbletea.(*Program).ReleaseTerminal()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tea.go:687 +0x164
  github.com/charmbracelet/bubbletea/teatest.TestModel()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/teatest/teatest.go:44 +0x378
  github.com/purpleclay/dns53/internal/tui_test.TestDashboard()
      /Users/paulthomas/dev/github.com/purpleclay/dns53/internal/tui/dashboard_test.go:50 +0x1dc
  github.com/purpleclay/dns53/internal/tui_test.TestDashboard()
      /Users/paulthomas/dev/github.com/purpleclay/dns53/internal/tui/dashboard_test.go:42 +0x19b
  testing.tRunner()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1439 +0x213
  testing.(*T).Run.func1()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1486 +0x47

Goroutine 8 (running) created at:
  github.com/charmbracelet/bubbletea/teatest.TestModel()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/teatest/teatest.go:34 +0x2fd
  github.com/purpleclay/dns53/internal/tui_test.TestDashboard()
      /Users/paulthomas/dev/github.com/purpleclay/dns53/internal/tui/dashboard_test.go:50 +0x1dc
  github.com/purpleclay/dns53/internal/tui_test.TestDashboard()
      /Users/paulthomas/dev/github.com/purpleclay/dns53/internal/tui/dashboard_test.go:42 +0x19b
  testing.tRunner()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1439 +0x213
  testing.(*T).Run.func1()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1486 +0x47

Goroutine 7 (running) created at:
  testing.(*T).Run()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1486 +0x724
  testing.runTests.func1()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1839 +0x99
  testing.tRunner()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1439 +0x213
  testing.runTests()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1837 +0x7e4
  testing.(*M).Run()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1719 +0xa71
  main.main()
      _testmain.go:99 +0x3a9
==================
==================
WARNING: DATA RACE
Read at 0x00c0003b7e90 by goroutine 8:
  bytes.(*Buffer).tryGrowByReslice()
      /usr/local/Cellar/go/1.18.3/libexec/src/bytes/buffer.go:107 +0x52
  bytes.(*Buffer).Write()
      /usr/local/Cellar/go/1.18.3/libexec/src/bytes/buffer.go:170 +0x18
  fmt.Fprintf()
      /usr/local/Cellar/go/1.18.3/libexec/src/fmt/print.go:205 +0xb1
  github.com/charmbracelet/bubbletea.clearLine()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/screen.go:19 +0x85
  github.com/charmbracelet/bubbletea.(*standardRenderer).stop()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/standard_renderer.go:75 +0x31
  github.com/charmbracelet/bubbletea.(*Program).shutdown()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tea.go:588 +0xd1
  github.com/charmbracelet/bubbletea.(*Program).StartReturningModel()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tea.go:496 +0x1cd1
  github.com/charmbracelet/bubbletea.(*Program).Start()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tea.go:549 +0x52
  github.com/charmbracelet/bubbletea/teatest.TestModel.func1()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/teatest/teatest.go:35 +0x53

Previous write at 0x00c0003b7e90 by goroutine 7:
  bytes.(*Buffer).tryGrowByReslice()
      /usr/local/Cellar/go/1.18.3/libexec/src/bytes/buffer.go:108 +0xb3
  bytes.(*Buffer).Write()
      /usr/local/Cellar/go/1.18.3/libexec/src/bytes/buffer.go:170 +0x18
  fmt.Fprintf()
      /usr/local/Cellar/go/1.18.3/libexec/src/fmt/print.go:205 +0xb1
  github.com/charmbracelet/bubbletea.showCursor()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/screen.go:15 +0x4d
  github.com/charmbracelet/bubbletea.Program.restoreTerminalState()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tty.go:30 +0x25
  github.com/charmbracelet/bubbletea.(*Program).ReleaseTerminal()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tea.go:687 +0x164
  github.com/charmbracelet/bubbletea/teatest.TestModel()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/teatest/teatest.go:44 +0x378
  github.com/purpleclay/dns53/internal/tui_test.TestDashboard()
      /Users/paulthomas/dev/github.com/purpleclay/dns53/internal/tui/dashboard_test.go:50 +0x1dc
  github.com/purpleclay/dns53/internal/tui_test.TestDashboard()
      /Users/paulthomas/dev/github.com/purpleclay/dns53/internal/tui/dashboard_test.go:42 +0x19b
  testing.tRunner()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1439 +0x213
  testing.(*T).Run.func1()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1486 +0x47

Goroutine 8 (running) created at:
  github.com/charmbracelet/bubbletea/teatest.TestModel()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/teatest/teatest.go:34 +0x2fd
  github.com/purpleclay/dns53/internal/tui_test.TestDashboard()
      /Users/paulthomas/dev/github.com/purpleclay/dns53/internal/tui/dashboard_test.go:50 +0x1dc
  github.com/purpleclay/dns53/internal/tui_test.TestDashboard()
      /Users/paulthomas/dev/github.com/purpleclay/dns53/internal/tui/dashboard_test.go:42 +0x19b
  testing.tRunner()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1439 +0x213
  testing.(*T).Run.func1()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1486 +0x47

Goroutine 7 (running) created at:
  testing.(*T).Run()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1486 +0x724
  testing.runTests.func1()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1839 +0x99
  testing.tRunner()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1439 +0x213
  testing.runTests()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1837 +0x7e4
  testing.(*M).Run()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1719 +0xa71
  main.main()
      _testmain.go:99 +0x3a9
==================
==================
WARNING: DATA RACE
Read at 0x00c000156ce8 by goroutine 8:
  runtime.racereadrange()
      <autogenerated>:1 +0x1b
  github.com/charmbracelet/bubbletea.(*Program).StartReturningModel()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tea.go:496 +0x1cd1
  github.com/charmbracelet/bubbletea.(*Program).Start()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tea.go:549 +0x52
  github.com/charmbracelet/bubbletea/teatest.TestModel.func1()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/teatest/teatest.go:35 +0x53

Previous write at 0x00c000156ce9 by goroutine 7:
  github.com/charmbracelet/bubbletea.(*Program).ReleaseTerminal()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/tea.go:682 +0xc4
  github.com/charmbracelet/bubbletea/teatest.TestModel()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/teatest/teatest.go:44 +0x378
  github.com/purpleclay/dns53/internal/tui_test.TestDashboard()
      /Users/paulthomas/dev/github.com/purpleclay/dns53/internal/tui/dashboard_test.go:50 +0x1dc
  github.com/purpleclay/dns53/internal/tui_test.TestDashboard()
      /Users/paulthomas/dev/github.com/purpleclay/dns53/internal/tui/dashboard_test.go:42 +0x19b
  testing.tRunner()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1439 +0x213
  testing.(*T).Run.func1()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1486 +0x47

Goroutine 8 (running) created at:
  github.com/charmbracelet/bubbletea/teatest.TestModel()
      /Users/paulthomas/go/pkg/mod/github.com/charmbracelet/bubbletea@v0.21.1-0.20220621133951-26d32cd5beb0/teatest/teatest.go:34 +0x2fd
  github.com/purpleclay/dns53/internal/tui_test.TestDashboard()
      /Users/paulthomas/dev/github.com/purpleclay/dns53/internal/tui/dashboard_test.go:50 +0x1dc
  github.com/purpleclay/dns53/internal/tui_test.TestDashboard()
      /Users/paulthomas/dev/github.com/purpleclay/dns53/internal/tui/dashboard_test.go:42 +0x19b
  testing.tRunner()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1439 +0x213
  testing.(*T).Run.func1()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1486 +0x47

Goroutine 7 (running) created at:
  testing.(*T).Run()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1486 +0x724
  testing.runTests.func1()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1839 +0x99
  testing.tRunner()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1439 +0x213
  testing.runTests()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1837 +0x7e4
  testing.(*M).Run()
      /usr/local/Cellar/go/1.18.3/libexec/src/testing/testing.go:1719 +0xa71
  main.main()
      _testmain.go:99 +0x3a9
==================
--- FAIL: TestDashboard (0.21s)
    testing.go:1312: race detected during execution of test
```

caarlos0:
ahh, glad you got to test it @purpleclay!



> 1. My `Update()` method is driven by the initial `Init()` and needs it to return a `tea.Msg` containing EC2 metadata. This requires me to add an enforced sleep. Is this something most people will have to do? As I noticed it in your `simple` example that you do this also. Is there not a way for `teatest` to wait for all `Init` commands to complete automatically?

in that example case, the sleep is to ensure some time passed so I can properly test the output... that said, since commands run in another goroutine, it might be needed to either sleep or sync with a `chan` or something like that...

> 2. The generation of golden files through the `-update` flag is a nice touch. Would it feel more intuitive to do this through a `go:generate`?

I think you can put a `//go:generate go test -update` where you need... 

> 3. Do you envisage people testing partial parts of the terminal output within `teatest.TestModel(func(out []byte) {})`? I imagine I would always use the golden file approach. Attempting to assert against fragments of text would require me to include special characters e.g. `  �[1;;mPHZ:�[0m             AAAAAAAAAAAAAAAAAAAAAAAA [testing] `. Partial matching would be useful if these characters could be stripped from the output

I think some people might, for simpler outputs...

> 4. It seems you cannot run any of the `teatest` test cases if you provide the `-race` flag to `go test`. Is that by design? This is the current output from my test:

investigating...

> 5. I am thinking of including a ticker in my TUI. I am wondering what impact this would have on the golden file generated. Would we get sporadic failures

depending on your project, yes... unfortunately.. I don't have a good solution for that... but I'm open to hear any ideas you might have...

caarlos0:
@purpleclay the race should be fixed now!

purpleclay:
> @purpleclay the race should be fixed now!

Awesome. I will give it another try

examples/simple/main_test.go
```diff
@@ -0,0 +1,34 @@
...
+				func(p teatest.Program, in io.Writer) {
+					time.Sleep(time.Second + time.Millisecond*200)
+					p.Send("ignored msg")
+					p.Send(tea.KeyMsg{
+						Type: tea.KeyEnter,
+					})
+				},
+				func(out []byte) {
+					teatest.RequireEqualOutput(t, out)
+				},
```
twpayne, line 30: Firstly, thank you for this PR! I'm currently building my first bubbles and want to test them. Here's some feedback based on this PR based on my own testing efforts in https://github.com/twpayne/chezmoi/pull/2359.

At this point, I'm testing that my models have certain properties after receiving a bunch of messages, so I'd like to include that in the testing framework. A specific suggestion would be to make `teatest.Model` accept [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) (just like most of bubbletea already does), so I can write code like:

```go
    teatest.Model(
        t, m,
        teatest.WithSender(p teatest.Program, in io.Writer) {
            // as above
        },
        teatest.WithRequireOutput(/* ... */),
        teatest.WithValidateFinalModel(func(m tea.Model) err {
            // check some properties of m
        },
    )
```

teatest/teatest.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// TypeText types the given text into the given program.
+func TypeText(p Program, s string) {
+	for _, c := range []byte(s) {
+		p.Send(tea.KeyMsg{
+			Runes: []rune{rune(c)},
+			Type:  tea.KeyRunes,
+		})
+	}
```
twpayne, line 71: It's would also be nice to be able to send `tea.KeyMsg`s of different types, e.g. for `tea.KeyCtrlC`, etc. There's an initial implementation at https://github.com/twpayne/chezmoi/blob/4f4760f4911ff03fe431ce31bd847a05b213d6a9/pkg/chezmoibubbles/chezmoibubbles_test.go. This converts a string like `"a\r"` to the equivalent of the user pressing `a` then `Enter`.

twpayne (commented):
Thank you for this!

I'm currently writing tests for a few simple models that build on `github.com/charmbracelet/bubbles/textinput.Model` (code in https://github.com/twpayne/chezmoi/pull/2359).

I would like to be able to test for when `tea.Model.Update()` returns `tea.Quit` as the command, so that I know that my model has terminated correctly.  This does not seem to be currently possible as `tea.Cmd` is a function. Go only allows for functions to be compared to `nil`, and `tea.Quit` uses an [unexported struct value](https://github.com/charmbracelet/bubbletea/blob/a7d6b5cd70f1cf63855967a20627f43b663904dd/tea.go#L155-L157) to [signal its value](https://github.com/charmbracelet/bubbletea/blob/a7d6b5cd70f1cf63855967a20627f43b663904dd/tea.go#L496-L500). Would it be possible to add something like:

```go
    func IsCmdQuit(cmd tea.Cmd) bool { /* ... *. }
```

to `teatest` so `cmd.Quit`s can be tested?

examples/simple/main_test.go
caarlos0: That sounds like a good idea, yes. Will try to do something later today 🙏

teatest/teatest.go
caarlos0: That seems to make sense as well, will play with it later today 🙏

caarlos0:
I think it's possible, yes. Will work on this later 🤘

purpleclay:
@caarlos0 is there a potential timeframe of when this `teatest` package will make it into the main branch?

caarlos0:
> @caarlos0 is there a potential timeframe of when this `teatest` package will make it into the main branch?

no specific timelines yet, no :(

erwin:
When running tests with `teatest` what will be the terminal characteristics, rows and columns for the test terminal? Inside of `teatest.go` and `examples/simple/main_test.go` I didn't see a way to specify the terminal height, width, colordepth, or emulator name. Would a `tea.WindowSizeMsg` need to be sent? 

Should we have a basic one that's always sent to tests by default?

teatest.go -> `TypeText()` function looks very useful!

What do you think of a `func RequireMatchRegexp` in teatest.go that could be called multiple times to make sure that certain Regexes do match and other Regexes fail to match? Seems like that could a lot more flexible than matching a file output exactly, especially if the terminal dimensions may change.

On the `test` branch, using `./examples/simple` when I run `go test -v`, for all 3 of the instances I'm getting: (where X = 1, 2, 3)

```
=== CONT  TestApp/X
    main_test.go:30: output does not match, diff:
        
        1,2c1,2
        Hi. This program will exit in 9 seconds. To quit sooner press any key.
        ---
        Hi. This program will exit in 9 seconds. To quit sooner press any key.
```

Not sure what is causing the example to fail for me.

After running: `go test -update` (which passes) and running `go test -v` again I get:

```
=== CONT  TestApp/0
    main_test.go:30: output does not match, diff:
        
        3c3

        \ No newline at end of file
        ---

        \ No newline at end of file
=== CONT  TestApp/2
    main_test.go:30: output does not match, diff:
        
        3c3

        \ No newline at end of file
        ---

        \ No newline at end of file
```

erwin:
What do you think of adding a `RequireRegexpOutput` to `teatest`?

```
        teatest.WithRequiredOutputChecker(func(out []byte) {
          teatest.RequireRegexpOutput(t, out, `Hi\. This program will exit in \d seconds\. To quit sooner press any key\.`)
        }),

```

Then inside of teatest.go just add something like:

```
func RequireRegexpOutput(tb testing.TB, out[]byte, re string) {
  rexp, err := regexp.Compile(re)
  if err != nil {
    tb.Fatalf("Error Compiling Regexp: %s", err)
  }
  match := rexp.Match(out)
  if ! match {
    tb.Fatalf("output does not match:\nRegExp: %s\nOutput: %s", rexp.String(), string(out))
  }
}
```

Over time could improve that to help people narrow it down to exactly the portion of the RegExp that caused the match to fail. I'll have to do some research on that.

erwin:
I noticed when using `teatest.withProgramInstructions()` and sending `tea.WindowSizeMsg`, the test will hang indefinitely.

```
   p.Send("ignored msg")
>> p.Send(tea.WindowSizeMsg{ Height: 24 })
   p.Send(tea.KeyMsg{ Type: tea.KeyEnter, })
```

caarlos0:
> Personally, I would like to test tea components at various sizes and know that the planned information appeared on screen at each size.

fwiw added a `WithInitialTermSize` option

muesli:
Related work: https://github.com/charmbracelet/bubbletea/pull/536

This provides a generic event filter that we can use to inspect events in tests. It also exports `QuitMsg` so you can identify such messages.

caarlos0:
That looks awesome @muesli !

erwin:
In `examples/simple/main_test.go` on the test branch, if the model is initialized to count down for `10` seconds, and then sleeps for 1.2 seconds on [line 19](https://github.com/charmbracelet/bubbletea/blob/40565724aea200bea675613b3f793000c6e323d6/examples/simple/main_test.go#L19), then shouldn't we be testing to make sure the value of the model has been decremented from 10 to 9 on [line 31](https://github.com/charmbracelet/bubbletea/blob/40565724aea200bea675613b3f793000c6e323d6/examples/simple/main_test.go#L31)?

My apologies if I'm wrong and just misunderstanding the intention here.

If we should be testing for `9` rather than `10` here, then in teatest -> `TestModel` do we need get a copy of the `Update()`ed model back after `tea.Run` [before running `opts.validateModel`](https://github.com/charmbracelet/bubbletea/blob/40565724aea200bea675613b3f793000c6e323d6/teatest/teatest.go#L118
) that we set with `teatest.WithValidateFinalModel`?

caarlos0:
ah, indeed, good catch @erwin !

fixed on last commit

erwin:
I was writing some tests today and noticed cases where I should make sure that a RegEx does not match. For this particular example, I have an array assigned to the viewport bubble, and based on the size and keyboard input, some items should be present, and others not present.

Your `teatest.WithRequiredRegexpOutput` works great to test for output you want present, but if the match fails, it will call `t.Fatal` so you can't easily invert it to test for !WithRequiredRegexpOutput

@caarlos0 What do you think of another helper in `teatest` for cases where you don't want the regexp to match?

caarlos0:
@erwin tbh I'm thinking about removing the regex one completely... you already have the `testing.T` and the output, so it just does the regex matching... which many libs (like testify) already provide helpers for...

IMHO teatest should help setting up the app for testing, but that does not include implemente a test matchers library 🤔

erwin:
The reason I thought many convenience functions like regex wrappers (and any other widely useful ones) makes sense as part of `teatest` is because you already have what I perceived as the rather unusual (very special case?) `RequireEqualOutput` function, and something like string matching seems more generally useful.

I think functions that auto strip escape sequences from `out` also come to mind as useful.

The Regex helper is very trivial code, and as you point out, `out` is exposed, so people can easily implement whatever they want. I don't have any deep feeling about whether that is in or out of teatest. Since not everyone loves Regex's, perhaps direct string matching stuff would be even more generally useful.

I suppose my point is that the testing library and the corresponding example code both SHOUT VERY LOUDLY to new developers as they first adopt bubbletea. Every developer that uses bubbletea will experiment with at least some of the examples. The smoother and cooler that experience, the more bubbletea and charm.sh can grow.

I think that the `RequireEqualOutput` function comes across as too strange, and maybe just needs a comment to sell why it's such a great way to test. Or if you think everybody should be using `testify` in their tests, maybe some more in the examples that show why `testify` is so valuable. 

So far, `teatest` is very useful to me and I am very glad that you've written it!

caarlos0:
@erwin thanks for the feedback! 

`RequireEqualOutput` does a few more things, more related to golden file testing: create/update the .golden files, compare with diff, etc... maybe that could/should be in another repository?

erwin:
> maybe that could/should be in another repository?

That's an interesting idea! Or maybe just another package name, depending on how you anticipate it growing!

aschey:
I come from a webdev/React background and we typically use [React Testing Library](https://testing-library.com/docs/react-testing-library/example-intro) to test our UIs, so naturally, I was looking for a similar approach to test my TUI apps. I've been experimenting with [this small package](https://github.com/aschey/tui-tester) built on top of [termtest](https://github.com/activestate/termtest) to achieve this. The idea is pretty similar to this approach, but instead of sending a sequence of commands and then doing an assert at the end, the flow is more like
```
loadApp()
send("some input")
assert(app.output.contains("some output"))

send("some more input")
assert(app.output.contains("some more output"))
```

For more complex apps, I tend to prefer fine-grained asserts rather than matching the entire output so that all your tests don't fail as soon as you change one small part of your UI. Snapshot testing in React has somewhat fallen out of favor for this reason.

My app also heavily relies on styling to show UI changes so being able to do asserts on the foreground/background colors after each interaction is quite important - something like `output(row, col).background == green`.

Not saying this approach is better or worse, just some other ideas to consider.

caarlos0:
@aschey ... that's an interesting concept... will experiment with it a bit

caarlos0:
okay, I think we can have both options (current and the new proposed), will push an example in a few

aschey:
Nice, looks good! Since UI updates happen asynchronously, most UI testing libraries have a default timeout for any assertions so you don't have to hardcode sleeps in your program. Something like 
```go
err := p.WaitFor(func (out []byte) bool { 
   return bytes.Contains(out, []byte("This program will exit in 9 seconds"))
}, 2 * time.Second)
```
which will wait for up to 2 seconds before returning an error if the condition fails might be useful. Of course users could write this function themselves, but I think it would be so commonly used that having it baked in might be beneficial.

caarlos0:
@aschey had to add a couple more params, but just pushed a version of it... let me know what you think

aschey:
Awesome, thanks! Only thing I'd mention is the extra parameter for the polling interval since I think the vast majority of people won't care about that. Most libraries I've used set it to something small like 50 ms by default. It could be an option like `WithCheckInterval` so it's still configurable if you need it, but it avoids having to pass the extra parameter each time. On the other hand, it's not difficult to just make your own wrapper function on top of `WaitFor`, so I'm cool with either way.

The last thing I'd mention is the overall structure of having all program interactions in a callback might make this a bit inconvenient to use with some testing frameworks that require control over the structure of your tests. For example, I like to use [ginkgo](https://github.com/onsi/ginkgo) to organize integration tests with multiple steps. If `teatest.TestModel` returned some kind of wrapper on top of `tea.Program` instead, that might make it a bit more flexible. An example test would look something like:

```go
var _ = Describe("Tests", func() {
    m := model(10)
    t := GinkgoT()
    p := teatest.TestModel(t, m, teatest.WithInitialTermSize(70, 30))

    When("the user enters some text", Ordered, func() {
        BeforeAll(func() {
	    p.Send("text")
        })

        It("updates the view", func() {
            p.WaitFor(func(out []byte) bool {
                return bytes.Contains(out, []byte("text"))
            }, 5*time.Second, time.Second)
        })

        When("the program finishes", Ordered, func() {
            BeforeAll(func() {
                p.Quit()
                p.Wait()
            })
            
            It("renders the final view", func() {
                p.CheckOutput(func(out []byte) {
                    teatest.RequireEqualOutput(t, out)
                })
            })

            It("updates the model", func() {
                p.ValidateFinalModel(func(mm tea.Model) error {
                    m := mm.(model)
                    if m != 9 {
                        return fmt.Errorf("expected model to be 9, was %d", m)
                    }
                    return nil
                })
            })
        })
    })
})
```

caarlos0:
@aschey yeah, I think your idea is a good one, I'll look into how it would like implementation-wise as well

KarolosLykos:
Hello there!

I noticed this pull request and I'm wondering if it's still in development? Can you give us an update or any news on the progress?

Thank you!

teatest/teatest.go
```diff
@@ -0,0 +1,262 @@
...
+	if err := os.WriteFile(path, out, 0o600); err != nil { // nolint: gomnd
+		tb.Fatal(err)
+	}
+
+	// inspired by https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/cmd/internal/diff/diff.go;l=18
+	diff, err := exec.Command(
+		"diff",
+		"--minimal",
+		"--ignore-trailing-space",
```
bashbunni, line 238: this is causing an error on mine when running `examples/simple`
```
=== RUN   TestApp
    main_test.go:41: output does not match, diff:
        
        diff: unrecognized option `--ignore-trailing-space'
        diff: Try `diff --help' for more information.
--- FAIL: TestApp (1.22s)
=== RUN   TestAppInteractive
--- PASS: TestAppInteractive (3.07s)
FAIL
FAIL	examples/simple	4.292s
FAIL
```

deleting this line fixed it, but not sure if it caused other side effects elsewhere yet.

caarlos0: what's the output of `diff --version`?

bashbunni: `diff (GNU diffutils) 2.8.1`

caarlos0: hmmm mine is `diff (GNU diffutils) 3.8`... that explains it. Will look into it

bashbunni: Kk let me know if there's anything I can help with

twpayne: To avoid relying on external diff tools, consider using an existing Go package to implement diffs. github.com/sergi/go-diff and github.com/pkg/diff are popular.

caarlos0:
Hey everyone!   After many months, we're finally releasing the teatest package! 

It'll live under [github.com/charmbracelet/x/exp/teatest](http://github.com/charmbracelet/x/exp/teatest), and will have a tagged release once we release bubbletea.

 Please, do try it out in your projects, and open issues if you find any bugs or possible enhancements.

This PR right now contains only an example of teatest usage.

Thank you!

PR #356, opened 2022-06-27
added test for rune error

LukaJaj:
I added test for rune decoding error. is it ok way to mock ?


muesli:
I have no issue with mocking, but I'm wondering if this is really required here. Couldn't we just supply `readInputs` with an actual invalid input?

LukaJaj:
Correct, but reason we need to mock is utf8.DecodeRune() function since it requires utf8 (external) package

PR #363, opened 2022-07-05
Bump Bubbles to v0.13.0

maaslalani:
Upgrades `bubbles` to latest release.


maaslalani:
This affects only the `examples/` directory so that we include the textarea release.

PR #376, opened 2022-07-19
chore: fix Println typo

bashbunni:
## Changes:
- [x] change `Printf`to `Println` in the Godoc of the `Println` function

### DIFF
--- a/standard_renderer.go
+++ b/standard_renderer.go
@@ -483,10 +483,10 @@ type printLineMessage struct {
 	messageBody string
 }
 
-// Printf prints above the Program. This output is unmanaged by the program and
+// Println prints above the Program. This output is unmanaged by the program and
 // will persist across renders by the Program.
 //
-// Unlike fmt.Printf (but similar to log.Printf) the message will be print on
+// Unlike fmt.Println (but similar to log.Println) the message will be print on
 // its own line.
 //
 // If the altscreen is active no output will be printed.


muesli (approved):

maaslalani (approved):

standard_renderer.go
```diff
@@ -483,10 +483,10 @@ type printLineMessage struct {
 	messageBody string
 }
 
-// Printf prints above the Program. This output is unmanaged by the program and
+// Println prints above the Program. This output is unmanaged by the program and
 // will persist across renders by the Program.
 //
-// Unlike fmt.Printf (but similar to log.Printf) the message will be print on
+// Unlike fmt.Println (but similar to log.Println) the message will be print on
```
meowgorithm, line 489: I think we can actually drop this part of the comment altogether because `fmt.Println`, `log.Println`, and `tea.Println` all print on their own lines.

muesli: They finish with a line feed, but they wouldn't *start* on their own line, right? That's how I understood this comment anyway.

meowgorithm: I _think _ they all start on their own lines as well (unless I’m totally missing something, which is not unlikely)

muesli: Oh, they don't: they'll just start printing at the current cursor position, wherever that may be. Example:

```go
fmt.Print("Hello ")
fmt.Println("World")
```

meowgorithm: TIL. I stand corrected!

PR #395, opened 2022-08-11
docs: README progress-download folder + ignore

JeromeFitz:
📝 Update folder name in README to `progress-download` as
 that is what generates with `go build .`

🙈 Add the files generated via README to root `.gitignore`


maaslalani:
Thank you for this fix!

PR #396, opened 2022-08-12
fix: key bindings for urxvt, add home/end combinations

### LINKED ISSUE #310: capture ctrl+Home/End keys
It would be great if we can capture ctrl+Home/End key combinations. This work most probably will depend on improvements from #293

### LINKED ISSUE #403: Input key codes for home/end are incorrect
The code parses `CSI 1 ~` and `CSI 4 ~` for Home/End. This is incorrect. Should be 7/8 (for rxvt) and H/F for vt100/everything else.
Fixed by #396

knz:
Fixes #310.
Fixes #403.

cc @muesli


knz:
cc @maaslalani

meowgorithm:
Let's make sure we verify this one very carefully, including in both `urxvt` and otherwise.

knz:
All right I think there are two different changes here. I will split the PR which will simplify.

knz:
Ready for another look - I have split up the PR into bite-sized commits, so that they become easier to review individually.

muesli:
@knz Could you do me a favor and rebase this once more?

knz:
Done.

key.go
```diff
@@ -382,10 +382,10 @@ var sequences = map[string]Key{
 	"\x1b[2;3~":   {Type: KeyInsert, Alt: true},
 	"\x1b[3~":     {Type: KeyDelete},
 	"\x1b[3;3~":   {Type: KeyDelete, Alt: true},
-	"\x1b[1~":     {Type: KeyHome},
```
muesli, line 385: While you're right and most every terminal seems to send `\x1b[H` for the home key, I do find references to this sequence being as home as well. We should probably keep it?

muesli: Same question/remark for `KeyEnd`, too.

knz: On which terminal was that?

muesli: I'll update this commit and keep the previous sequences for `KeyHome` and `KeyEnd` for now. If we can confirm they aren't used anywhere we can still remove them eventually.

key.go
```diff
@@ -377,27 +377,32 @@ var sequences = map[string]Key{
...
-	"\x1b\x1b[2~": {Type: KeyInsert, Alt: true}, // urxvt
-	"\x1b\x1b[3~": {Type: KeyDelete, Alt: true}, // urxvt
-	"\x1b\x1b[5~": {Type: KeyPgUp, Alt: true},   // urxvt
 	"\x1b\x1b[6~": {Type: KeyPgDown, Alt: true}, // urxvt
-	"\x1b\x1b[7~": {Type: KeyHome, Alt: true},   // urxvt
-	"\x1b\x1b[8~": {Type: KeyEnd, Alt: true},    // urxvt
+
+	"\x1b[H":    {Type: KeyHome},            // vt100
+	"\x1b[1;3H": {Type: KeyHome, Alt: true}, // vt100
+	"\x1b[F":    {Type: KeyHome},            // vt100
```
muesli: I think this should read `KeyEnd`. Fixing.

muesli:
@knz Help me understand why the last commit improves troubleshootability? I find the function keys a bit all over the place still.

key.go
muesli: Ok, pushed my changes to your fork.

key.go
muesli: Fixed and pushed, but noticed you corrected the mistake yourself in one of the later commits.

key.go
knz: Ok this works for me. thanks for the edit!

knz:
> why the last commit improves troubleshootability?

As I was cross-checking the codes for the function keys with the source code of all the terminal emulators listed in the termenv keys, I was doing them in order of function key (starting at F1, then F2, etc). It was super hard to do because they were not in function key order in key.go. That's why I rearranged them. But feel free to drop that latest commit if you believe it's not worth the hassle.

muesli:
Thank you @knz, great work 🙌

PR #397, opened 2022-08-14
feat: bracketed paste

knz:
First 3 commits from #568, #569 and #570.
Partial fix to #404

This introduces support for input via bracketed paste, where escape
characters in the pasted input are not interpreted.
Pasted input are marked as a special field in the KeyMsg.
This is because pasted input may need sanitation in individual widgets.

cc @maaslalani @muesli

### DIFF
--- a/key.go
+++ b/key.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"regexp"
+	"strings"
 	"unicode/utf8"
 )
 
@@ -54,6 +55,7 @@ type Key struct {
 	Type  KeyType
 	Runes []rune
 	Alt   bool
+	Paste bool
 }
 
 // String returns a friendly string representation for a key. It's safe (and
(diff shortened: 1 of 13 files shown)


knz:
As discussed with @muesli I changed the PR to always enable bracked paste, unless the option WithoutBracketedPaste is specified on the program.

key.go
```diff
@@ -565,3 +634,43 @@ func readInputs(input io.Reader) ([]Msg, error) {
...
 	return msgs, nil
 }
+
+// SanitizeRunes removes control characters from runes in a KeyRunes
+// message, and optionally replaces newline/carriage return by a
+// specified character.
+//
+// The rune array is modified in-place. The returned slice
+// is the original slice shortened after the control characters have been removed.
+func SanitizeRunes(runes []rune, replaceNewLine rune) []rune {
```
meowgorithm, line 644: Is there any particular reason this function needs to be exported?

knz:
Rebased to avoid conflict.

muesli:
Rebasing.

knz:
I think this code will be simplified / become clearer if we implement it on top of #511.

knz:
@muesli this is ready again.

knz:
@muesli in case you haven't found it already, this patch has a shortcoming: certain terminal emulators feed the pasted datas as non-continuous writes to the terminal buffer, so the bubbletea code can sometimes read less than `len(buf)` before the paste end marker is encountered.

I will iterate and submit a better version.

knz:
I have added a commit on top of the original branch. @muesli you can look at just the diff for the latest commit to see what you need to change on your branch.

xprnio:
Bumping. I have a use-case similar to #800 and would love to be able to detect `ctrl+enter`, but AFAIK this PR is what would solve that.

meowgorithm:
@xprnio actually, what you want for shift+enter is #869. Will also comment on that issue.

nadimkobeissi:
Please merge this!

PR #412, opened 2022-08-20
Catch SIGTERM so that bubbletea cleans up terminal

DiabeticOwl:
Add [SIGTERM](https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html#index-SIGTERM) to signals caught so that bubbletea cleans up the screen when exiting because of that signal.

Unix utilities commonly send SIGTERM as the signal to terminate a process. Currently bubbletea doesn't catch SIGTERM and if a application receives this type of signal it will exit leaving the terminal in an unusable state. This PR causes this package to prevent this behavior and clean the terminal on exit. Examples of Unix utilities that send SIGTERM as default include `kill` and `timeout`.

To reproduce this bug run: `timeout 10 ./bubbletea-progam` 

I have tested this PR with both `timeout` and `kill`.

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -337,11 +337,14 @@ func (p *Program) StartReturningModel() (Model, error) {
 		p.input = f
 	}
 
-	// Listen for SIGINT and SIGTERM. Note that in most cases ^C will not send an
-	// interrupt because the terminal will be in raw mode and thus capture
-	// that keystroke and send it along to Program.Update. If input is not a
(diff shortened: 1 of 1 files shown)


meowgorithm:
LGTM. This is a good one—thank you!

raphaelfff:
How can I have some specific handling of a ctrl+C in that case ?

PR #415, opened 2022-08-24
feat: add `Sequence` command for running commands in order

### LINKED ISSUE #413: Using tea.Printf before tea.Quit
it is quite difficult to use the `tea.Printf` and `tea.Println` functions right before `tea.Quit`.
When using it with `tea.Batch` and `tea.Quit` for example, there is a race between `tea.Printf` and `tea.Quit`.

Example:

```go
// in this example, we are dealing with a list of tasks,
// display the latest state then start the next task.

  case foo:
     m.current++
     // if there is still work to do
     return m, tea.Batch( // this works fine
       tea.Println(m.tasks[m.current-1].View()),
       m.tasks[m.current].Update(),
     )
     // ...
     // Last element of the list, print it and leave the program
     return m, tea.Batch( // this doesn't work, race between Println and Quit
       tea.Println(m.tasks[m.current-1].View()),
       tea.Quit,
     )
```

By design `tea.Batch` doesn't provide any ordering guarantees, however its use with `tea.Printf` and `tea.Quit` makes it difficult to display *all* messages before quitting the program.

(If I understand correctly, the tea.Print* [example](https://github.com/charmbracelet/bubbletea/blob/master/examples/package-manager/main.go) has the same issue, although it misses the first entry in the list (it's more visible if you deactivate the randomness of the packages function))

I tried with `tea.Sequentially` but it returns the first non-nil message, this ignores the `tea.Quit` message.

Ideally I would like to use `tea.Printf`, then once it's rendered, use `tea.Quit`. Example:

```go
  return m, tea.QuitAfter(tea.Printf("...."))

```

What is the proper way of using `tea.Printf` and then `tea.Quit`?

meowgorithm:
Closes #413. Also note that this deprecates `Sequentially`.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -119,6 +119,8 @@ func Tick(d time.Duration, fn func(time.Time) Msg) Cmd {
 //	}
 //
 //	cmd := Sequentially(saveStateCmd, Quit)
+//
+// Deprecated: use Sequence instead.
 func Sequentially(cmds ...Cmd) Cmd {
 	return func() Msg {
 		for _, cmd := range cmds {
--- a/examples/sequence/main.go
+++ b/examples/sequence/main.go
@@ -1,0 +1,40 @@
+package main
+
+// A simple example illustrating how to run a series of commands in order.
+
+import (
+	"fmt"
+	"os"
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+type model struct{}
+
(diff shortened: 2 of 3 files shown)


caarlos0 (approved):
ahh, that looks good!

knz (approved):
oh great, this is very useful.

PR #418, opened 2022-08-25
Support the insert key

knz:
This can be used in e.g. textarea / textinput to enable "overwrite
mode".

cc @muesli


muesli:
Thanks again, @knz!

PR #421, opened 2022-08-25
feat: support a Clear command

### LINKED ISSUE #116: Alternate screen doesn't clear the terminal when running in GNU Screen
### Steps to reproduce

1. Run GNU Screen (I'm running it in an OSX Terminal)
2. In the bubbletea repo, `go run examples/fullscreen/main.go`

Expected result: Screen is cleared and you only see the countdown message

Actual result: Your terminal prompt, history, and whatever else is still visible alongside the output of the program.

However, once the program exits, the contents of the terminal are cleared.

This bug does not happen in tmux or in a bare Terminal.

### Screenshot

(I ran `ls` just before running the program, to illustrate how it writes over top of the existing terminal contents)

Screen Shot 2021-07-13 at 9 38 44 PM

### LINKED ISSUE #236: clear screen Cmd
I would like to clear top of screen from within my app (same as Ctrl+L in bash). 

It seems I need to print "\033[H\033[2J" (for nix) and initiate full redraw. I see internal `renderer.repaint` is only initiated from `WindowSizeMsg`:
```go
		case "ctrl+l":
			fmt.Print("\033[H\033[2J") // only nix
			w, h, _ := term.GetSize(0) // wouldnt work if out is set
			return m, func() tea.Msg { // hack to trigger redraw
				return tea.WindowSizeMsg{Width: w, Height: h}
			}
		}
```

Is there an easier way to do this? 
Would be nice to have built-in `tea.ClearScreen` cmd or at least some way to trigger full redraw so solution is less hacky and cross-platform. 

clear_screen

knz:
Fixes #236.
Fixes #116.

cc @muesli @meowgorithm

### DIFF
--- a/nil_renderer.go
+++ b/nil_renderer.go
@@ -7,6 +7,7 @@ func (n nilRenderer) stop()                   {}
 func (n nilRenderer) kill()                   {}
 func (n nilRenderer) write(v string)          {}
 func (n nilRenderer) repaint()                {}
+func (n nilRenderer) clearScreen()            {}
 func (n nilRenderer) altScreen() bool         { return false }
 func (n nilRenderer) enterAltScreen()         {}
 func (n nilRenderer) exitAltScreen()          {}
--- a/renderer.go
+++ b/renderer.go
@@ -21,6 +21,9 @@ type renderer interface {
 	// in succession.
 	repaint()
 
+	// Clears the terminal.
+	clearScreen()
+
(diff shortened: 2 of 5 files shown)


knz:
Added a fix for #116 too while I was at it.

knz:
I've rebased this and enhanced the commit messages for clarity. PTAL

PR #422, opened 2022-08-25
Fix 1-indexing when entering alt screen

knz:
cc @muesli


muesli:
Will rebase & merge.

muesli:
Good catch, thank you @knz 🙌

PR #426, opened 2022-08-27
docs: fix typo

kotajacob:
The previous example was wrong and would result in the following
compiler error:
```
error (type) is not an expression
```
Just a simple typo, I've changed it to `err` to correctly reference the
function scoped variable named `err`.


muesli:
Thank you!

PR #428, opened 2022-08-29
keys: properly support the alt modifier

### LINKED ISSUE #427: Alt key not working properly for control characters
Alt+Enter, Alt+Tab, Alt+Ctrl+C, etc do not work properly: they are returned as `KeyMsg{Type: KeyRunes, Alt: true}` instead of the proper `Type`.

Because of this bug, to define a key binding for KeyEnter one has to do this instead:

```go
  AlwaysComplete:  key.NewBinding(key.WithKeys("alt+\r")),
```

which is ugly and disappointing. I wish `alt+enter` would work instead. The patch in #428 fixes it.

knz:
Fixes  #427.

cc @muesli @meowgorithm

### DIFF
--- a/key.go
+++ b/key.go
@@ -543,28 +543,29 @@ func readInputs(input io.Reader) ([]Msg, error) {
 
 		// Is the alt key pressed? If so, the buffer will be prefixed with an
 		// escape.
+		alt := false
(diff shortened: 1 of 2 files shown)


knz:
Also added a unit test.

muesli:
Thank you, @knz!

PR #444, opened 2022-09-09
Fixes for examples/progress-download

anacrolix:
Some fixes I made while playing around with this example. Thanks!


muesli:
Simplified the example a bit further. If we can't parse the content length, we simply abort. Good enough for an example.

muesli:
Thank you, @anacrolix!

PR #452, opened 2022-09-18
fix: visible cursor in altscreen (#190)

### LINKED ISSUE #190: Cursor visible in `cmd.exe` on Windows when the altscreen is active
I've tested a bubbletea program on windows 11 and it seems that the cursor always stays visible in the left corner of the terminal. ~~In the windows terminal as well as in powershell.~~ It seems that newer versions of windows terminal have fixed this. Affected is only cmd.exe / powershell.

Londek:
The reasoning behind this hideCursor is behavior of AltScreen in cmd.exe; cmd stores two different cursor "states" for both AltScreen and main buffer so, we have to explicitly call hideCursor whenever we enter AltScreen. This should also fix cursor bugs on other platforms

Closes #190

### DIFF
--- a/examples/spinners/main.go
+++ b/examples/spinners/main.go
@@ -55,7 +55,7 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			return m, tea.Quit
 		case "h", "left":
 			m.index--
-			if m.index <= 0 {
+			if m.index < 0 {
 				m.index = len(spinners) - 1
 			}
 			m.resetSpinner()


muesli:
Apologies, I meant to rebase this, not close it. I didn't realize you created this PR from your fork's master branch, which causes this issue. I will recover your commit and push it in a separate PR in a minute. Thank you for your contribution!

muesli:
There we go: https://github.com/charmbracelet/bubbletea/pull/462

PR #454, opened 2022-09-21
fix: off-by-one error in spinner example

smlx:
Sorry for the noise in `go.mod` and `go.sum`. I had to `go get github.com/charmbracelet/bubbles/spinner` in order to run the example.


muesli:
I've rebased and cleaned up the noise in `go.mod`/`go.sum`. We have a separate set of module definitions in `examples/go.mod`, so you can avoid this issue by running the examples from within the directory.

muesli:
Thank you, @smlx!

PR #467, opened 2022-09-27
fix: properly skip over unrecognized CSI sequences.

### LINKED ISSUE #466: Unknown CSI inputs result in garbage key events
When a terminal CSI sequence is not recognized by bubbletea, it is not completely ignored; instead, bubbletea will match the initial escape byte and the character immediately afterwards as a single key with the alt modifier, then interpret the rest as regular key inputs. For example, ESC [ 1003 h  gets parsed as alt+1, then 0, then 0, then 3, then h.

Since the CSI sequences all have a regular structure, they could be parsed and ignored as a single unit.

(The definition is strict: a CSI sequence starts with `ESC [` and terminates with a byte value in the range 0x40 to 0x7E.)

knz:
Fixes #466.

Also includes a removal of now-unnecessary code.

cc @muesli

### DIFF
--- a/key.go
+++ b/key.go
@@ -624,6 +624,12 @@ func readInputs(input io.Reader) ([]Msg, error) {
 			continue
 		}
 
+		// Is this an unrecognized CSI sequence? If so, ignore it.
+		if len(runes) > 2 && runes[0] == 0x1b && (runes[1] == '[' ||
+			(len(runes) > 3 && runes[1] == 0x1b && runes[2] == '[')) {
+			continue
+		}
+
 		// Is the alt key pressed? If so, the buffer will be prefixed with an
 		// escape.
 		alt := false
--- a/key.go
+++ b/key.go
@@ -2,7 +2,6 @@ package tea
 
 import (
 	"errors"
-	"fmt"
 	"io"
 	"unicode/utf8"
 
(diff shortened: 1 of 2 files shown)


key.go
```diff
@@ -469,18 +468,12 @@ var sequences = map[string]Key{
 	"\x1b\x1b[32~": {Type: KeyF18, Alt: true},
 	"\x1b\x1b[33~": {Type: KeyF19, Alt: true},
 	"\x1b\x1b[34~": {Type: KeyF20, Alt: true},
-}
-
-// Hex code mappings.
-var hexes = map[string]Key{
-	"1b0d": {Type: KeyEnter, Alt: true},
-	"1b7f": {Type: KeyBackspace, Alt: true},
```
muesli, line 477: Just checking in if this was removed on purpose? Since you added a test for the two sequences above, I assume it's working fine tho.

knz: Yes as the commit message says this is now unneeded. It became obsolete when we fixed the alt key a few PRs ago.

meowgorithm: I’ll add that I’m thrilled to see those gone. Thanks, @knz.

muesli:
Rebased master onto your branch.

PR #470, opened 2022-09-29
Example GIFs

maaslalani:
Adds `examples/README.md` as well as individual `README`s in each example folder to show what the code will look like when it is run.


bashbunni (approved):
Awesome, this looks great! Thanks Maas

Issue #299, opened 2022-04-15
Improve godoc on `tea.Program.Start()`


NiloCK:
https://github.com/charmbracelet/bubbletea/blob/14e58aa1f92fa9bd756d9e739ce63a39e98a39ac/tea.go#L584

This is an important function, which will be called by every beginner user in every program.

- "Ignores the final model." is not a complete sentence, and I don't understand it anyhow.
- There is no indication that `Start` is blocking. I scratched my head quite a bit before replacing this:

```
if err := game.Start(); err != nil {
	fmt.Printf("error starting server game: %v", err)
	os.Exit(1)
}

// why isn't my connection receiving game updates and forwarding them?
for {
	bf := make([]byte, 5120)
		n, err := conn.Read(bf)
	if err != nil {
		panic(err)
	}

	var gs c4.GameState
	json.Unmarshal(bf[:n], &gs)
	game.Send(gs)
}
```

with this:

```
go func() {
	for {
		bf := make([]byte, 5120)

		n, err := conn.Read(bf)
		if err != nil {
			panic(err)
		}
		var gs c4.GameState
		json.Unmarshal(bf[:n], &gs)

		game.Send(gs)
	}
}()

if err := game.Start(); err != nil {
	fmt.Printf("error starting server game: %v", err)
	os.Exit(1)
}
```

See the godoc on [exec.Run](https://pkg.go.dev/os/exec#Cmd.Run) and [exec.Start](https://pkg.go.dev/os/exec#Cmd.Start) for examples of well communicated blocking / nonblocking.

muesli:
I agree, the naming is a bit unfortunate indeed and calling it `Program.Run` could have avoided this irritation.

Two things we can do, as I see it:

1. Explicitly document the blocking nature of Start in the godocs.
2. Introduce `Program.Run` and update the examples & documentation. Deprecate `Start` and have it become an alias of `Run` for the time being.

The first one is an obvious and easy fix, not sure about the second option. Might be the right thing to do tho.

inkel:
I took a stab at updating the docs in #479.

inkel:
I for one think that would be a good idea, however, it is a quite disruptive change, so maybe it should be planned carefully 🤔 

Right now the _statu quo_ is using `Start`, and if newcomers already have problems with it because of confusing documentation then I'd start there, and slowly move towards implementing `Run`; all existing documentation, YouTube tutorials, etc, use `Start` so introducing the change now will be even more confusing IMO.

[...]

PR #479, opened 2022-10-04
Update Program.Start docs to indicate it's a blocking operation

inkel:
This addresses one of the concerns in #299.

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -335,7 +335,9 @@ func (p *Program) eventLoop(model Model, cmds chan Cmd) (Model, error) {
 	}
 }
 
-// StartReturningModel initializes the program. Returns the final model.
+// StartReturningModel initializes the program and runs its event loops,
+// blocking until it gets terminated by either [Program.Quit], [Program.Kill],
+// or its signal handler. Returns the final model.
 func (p *Program) StartReturningModel() (Model, error) {
 	handlers := handlers{}
 	cmds := make(chan Cmd)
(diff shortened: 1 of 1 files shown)


tea.go
```diff
@@ -562,7 +562,7 @@ func (p *Program) StartReturningModel() (Model, error) {
 	}
 }
 
-// Start initializes the program. Ignores the final model.
+// Start initializes the program and waits for it to end.
```
muesli, line 565: I would emphasize that it does more than just initializing the program: it executes its event loop and waits for the program to be terminated by a `tea.Quit` message.

inkel: Makes sense. Let me fire up my editor.

inkel: How about now?

meowgorithm: Keep in mind that the program can also be terminated by `Program.Quit`, `SIGINT` and `SIGTERM`. On that note I might remove the `tea.Quit` detail and simply state that the method blocks.

muesli: Technically `Program.Quit` is just `tea.Quit` as a function, but you're right. Instead of removing it, I think pointing that out in the docs wouldn't be a bad idea either. `tea` setting up its own signal handler isn't exactly obvious and users should be made aware of it.

meowgorithm: That's a great point, especially now that GoDocs support linking.

muesli:
Rebased, squashed, and further extended the comment to mention the signal handler.

muesli:
Thank you, @inkel!

inkel:
Thank **you** for building the easiest package to build TUI applications in the Go landscape by far!!! I love it.

Issue #460, opened 2022-09-25
tea.Print[f/ln] does not wrap to a new line


Charliego3:
When using tea.Print[f/ln] the text is too long and the terminal does not wrap to a new line.
The same string, fmt.Println output is normal, but this problem occurs when tea.Println is used for output

<details>
  <summary>go env</summary>
  <pre>
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/charlie/Library/Caches/go-build"
GOENV="/Users/charlie/Library/Application Support/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/charlie/dev/environment/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/charlie/dev/environment/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/Cellar/go/1.19.1/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.19.1/libexec/pkg/tool/darwin_amd64"
GOVCS=""
GOVERSION="go1.19.1"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/charlie/dev/go/transerver/cli/go.mod"
GOWORK="/Users/charlie/dev/go/transerver/go.work"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch x86_64 -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/29/61lvvzb95pq774dthcv3rhnh0000gn/T/go-build1132523527=/tmp/go-build -gno-record-gcc-switches -fno-common"
</pre>
</details>

OS: macOS 12.6
Terminal: kitty / mac's own terminal

```go
type model struct {
	spinner spinner.Model
	done    bool
}

[...]

PR #490, opened 2022-10-06
Support multiline tea.Println() messages

Adjective-Object:
When logging something into the scrollback history of the standard renderer, the current approach temporarially increases the size of the UI by the number of logged messages. This has 2 issues:

- log lines are truncated according to the current width of the terminal UI. This truncation behaviour is desireable when writing into the TUI's controlled space, but is not desireable when Printf()ing messages into the scrollback, since the contents of the logged line will be truncated with no way to recover the history.
- User code pre-wrapping messages they want to print to the scrollback don't behave the way that you would want/expect log lines in a terminal to behave:
  - Resizing the terminal emulator does not reflow the lines to the new terminal's width
  - Manual wrapping breaks most terminal emulator's detection of clickable URLs and file paths

This change replaces that approach with a full clear of the previos UI + printing unwrapped messages into the terminal before the next render

this has a minor performance impact because it necessitates a full clear of the terminal UI on each Printf(). This breaks duplicate line detection between renders. However, the current approach will rarely hit duplicate line detection, since adding log messages to the front of the printout buffer offsets the new buffer. Therefore, the current line detection logic will only trigger when the buffer contains duplicate lines offset by the number of messages being logged in this render pass.

### DIFF
--- a/standard_renderer.go
+++ b/standard_renderer.go
@@ -173,19 +173,18 @@ func (r *standardRenderer) flush() {
 	skipLines := make(map[int]struct{})
 	flushQueuedMessages := len(r.queuedMessageLines) > 0 && !r.altScreenActive
 
-	// Add any queued messages to this render
-	if flushQueuedMessages {
-		newLines = append(r.queuedMessageLines, newLines...)
-		r.queuedMessageLines = []string{}
-	}
-
 	// Clear any lines we painted in the last render.
 	if r.linesRendered > 0 {
 		for i := r.linesRendered - 1; i > 0; i-- {
-			// If the number of lines we want to render hasn't increased and
-			// new line is the same as the old line we can skip rendering for
-			// this line as a performance optimization.
-			if (len(newLines) <= len(oldLines)) && (len(newLines) > i && len(oldLines) > i) && (newLines[i] == oldLines[i]) {
+			// if we are clearing queued messages, we want to clear all lines, since
+			// printing messages allows for native terminal word-wrap, we
+			// don't have control over the queued lines
+			if flushQueuedMessages {
+				out.ClearLine()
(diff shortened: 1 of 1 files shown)


meowgorithm:
Related: https://github.com/charmbracelet/bubbletea/issues/460

Adjective-Object:
Oh, I'm sorry! I didn't see it already had an issue someone had intended to work on for hacktoberfest!

Didn't mean to step on anyone's toes, I've been kicking the can on this for a while and finally made this change this week because I have a deadline coming up.

bashbunni:
@Adjective-Object no worries! Issues aren't reserved; it's first come, first serve

knz:
Could we make this wrap the output instead of truncating it?

Adjective-Object:
That's what the PR does.

Adjective-Object:
Oh you mean with wordwrap? That has undesireable side effects I describe in
the description:
- println text does not reflow when resizing the terminal emulator
- links and file paths that are manually wrapped partway through are not
recognized by most terminal emulators

different55:
Looks like this PR is still fresh enough to apply, is there anything I could do to help this along? I've been running into the issue of unmanaged output not wrapping in one of my projects recently.

meowgorithm:
@Adjective-Object if this is ready we can take a look and most likely merge it quickly. Let us know.

standard_renderer.go
```diff
@@ -671,6 +680,7 @@ func ScrollDown(newLines []string, topBoundary, bottomBoundary int) Cmd {
 
 type printLineMessage struct {
 	messageBody string
+	// if true, skips truncation of the message
```
Adjective-Object, line 683: stray edit should be removed

Adjective-Object:
@meowgorithm, we've been using this fork internally at my work for a while, so it seems to be stable enough! Looks like I have a stray comment leftover from rebasing but other than that it should be good to look at in its current state

PR #493, opened 2022-10-07
Render view before quitting

### LINKED ISSUE #274: `Program#Quit` skips final render
As noted in #273, `Program#Quit` does not allow for a final render, and it should.

michal-kralik:
Closes #274 

Consolidate behavior before quitting so that final render is not skipped when calling `p.Quit()`.

Current state:
- `tea.Quit()` renders the model before quitting
- `p.Quit()` does not render the model before quitting

**Notes**
There does not seem to be a way to flush the buffer before quitting.
Could there be a race-condition where the model is not rendered before quitting?
Happy to update the PR if maybe `flush` method shall be exposed in the `renderer` interface.

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -373,6 +373,7 @@ func (p *Program) StartReturningModel() (Model, error) {
 			// Handle special internal messages.
 			switch msg := msg.(type) {
 			case quitMsg:
+				p.renderer.write(model.View())
 				cancelContext()
 				waitForGoroutines(p.cancelReader.Cancel())
 				p.shutdown(false)


muesli:
Thank you for your contribution! This fix is appropriate, but I've just cleaned up the entire init, tear-down, and event-loop structure of tea, which obviously causes a merge conflict. I've moved the final render call to the `Run` function, on the same level where we trigger the initial render.

muesli:
Apologies should this cause you to lose a Hacktoberfest contribution - this obviously wasn't my intention :disappointed:

PR #496, opened 2022-10-07
fix(key): recover the insert key

knz:
We started supporting insert in #418, but then accidentally removed it during a rebase in #396. Oops.

:facepalm: 

cc @muesli @meowgorithm


muesli:
Well this is embarrassing, I reviewed/rebased this as carefully as possible 🙈 Thanks @knz!

Issue #497, opened 2022-10-07
bug: cannot suspend bubbletea process


knz:
In unix shells it's expected that a terminal program suspends itself when the user pressses Ctrl+Z (which sends the signal SIGTSTP to the process).

However, bubbletea disables this terminal feature which results in Ctrl+Z becoming ineffective.

In [bubbline](https://github.com/knz/bubbline), we're solving this by using [this key handler](https://github.com/knz/bubbline/blob/71919fc07ec06e1187a5a2acb28325ca7e31d04c/editline/editline.go#L825-L835).

However this is cumbersome, and should be handled inside bubbletea directly, perhaps controlled by a top level program option.

cc @muesli

knz:
NB: if you're interested I could pull the functionality out of bubbline into a PR for bubbletea. Let me know.

muesli:
Agreed, since we already have a default signal handler, we should probably strive to support `SIGSTP` as well.

knz:
> since we already have a default signal handler, we should probably strive to support SIGSTP as well.

It's the other way around: the process does not handle SIGTSTP itself (the kernel does). However, Ctrl+Z should *send* SIGTSTP to the process.

I'll send a PR.

muesli:
I'm aware of that, I just meant we are already providing "extra convenience" like a default signal handler, so it probably makes sense to complete the signal handling. On a slightly related note, I'm currently making the signal handler optional, and I'm wondering if the same option flag should also affect `SIGSTOP` / `SIGQUIT` handling then.

PR #499, opened 2022-10-07
fix: properly detect terminal size after exec

knz:
See commits for details. Found while investigating #497.

cc @muesli

### DIFF
--- a/signals_unix.go
+++ b/signals_unix.go
@@ -8,8 +8,6 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
-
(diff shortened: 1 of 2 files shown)


muesli:
We could also use `checkResize` to ensure an initial terminal size is known before the first render 🤔

muesli:
Rebasing.

muesli:
Superseded by #541 which is based on this PR, but goes a step further and fixes a race condition accessing the `Program`'s `ctx`. Thank you, once again, @knz!

PR #500, opened 2022-10-07
feat: suspend on ctrl+z, abort on ctrl+\

### LINKED ISSUE #497: bug: cannot suspend bubbletea process

### LINKED ISSUE #498: bug: cannot invoke the go runtime's default goroutine dump behavior
When something goes wrong in a go process (e.g. infinite loop, synchronization deadlock etc) it's useful to get a goroutine/stack dump using the key combination Ctrl+\, which sends SIGQUIT to the process.

That's also a nice way to get a breakpoint if the debugger is attached.

However, bubbline disables this terminal behavior and Ctrl+\ becomes ineffective.

In [bubbline](https://github.com/knz/bubbline), we're solving this by using [this key handler](https://github.com/knz/bubbline/blob/71919fc07ec06e1187a5a2acb28325ca7e31d04c/editline/editline.go#L815-L823).

However this is cumbersome, and should be handled inside bubbletea directly, perhaps controlled by a top level program option.

cc @muesli

knz:
First two commits from #541.

Fixes #497.
Fixes  #498.

cc @muesli


knz:
just rebased.

knz:
@muesli anything else I need to do here?

PR #506, opened 2022-10-07
fix: kill should result in Start returning an error

muesli:
We can also drop the kill channel and rely on our existing context channel.


knz (commented):
nice

PR #509, opened 2022-10-07
chore: rename StartReturningModel to Run

muesli:
Added aliases for the old functions, but deprecated them.


muesli:
I'll merge #479 before this one and update the docs once more.

Issue #404, opened 2022-08-18
Missing support for bracketed paste


knz:
Bracketed paste is an optional feature that modifies how the terminal handles a "paste from clipboard" action by a user.
(e.g. middle mouse click in xterm, ctrl+shift+v in vte, etc)

The feature must first be enabled by sending `CSI ? 2004 h` to the terminal. After bracketed paste is enabled, each pasted text gets enclosed between `CSI 200~` and `CSI 201~`.

The reason why this feature exist is well explained in [the wikipedia page](https://en.wikipedia.org/wiki/Bracketed-paste#Motivation):

 
![image](https://user-images.githubusercontent.com/642886/185366561-36f4963b-598d-4bc3-a612-2c7ce9743542.png)

Likewise, in [the Bubbline library](https://github.com/knz/bubbline), which we intend to use for CockroachDB's SQL shell, we also want to interpret a *manually* entered newline as a special "run" command, but we also want to give users the option to *paste* a SQL text containing newline characters without running them.

### Prototype implementation

These 3 PRs:

- [ ] termenv: https://github.com/muesli/termenv/pull/84
- [ ] bubbletea: https://github.com/charmbracelet/bubbletea/pull/397
- [ ] bubbles: https://github.com/charmbracelet/bubbles/pull/214

### Terminal support

| Terminal         |Bracketed Paste |
| ---------------- |:-------------: |
| alacritty        |        ✅        |
| foot             |        ✅        |
| kitty            |        ✅        |
| Konsole          |        ✅        |
| rxvt             |        ✅        |
| screen           |        ❌        |
| st               |        ✅        |
| tmux             |        ✅        |
| vte-based  |        ✅        |
| wezterm          |        ✅        |
| xterm            |        ✅        |
| Linux Console    |        ❌        |
| Apple Terminal   |        ✅        |
| iTerm            |        ✅        |
| Windows cmd      |        ❌        |
| Windows Terminal |        ✅        |

knz:
See public slack discussion here: https://charmbracelet.slack.com/archives/CQ2E9CGTV/p1660812723955889

knz:
There's a couple implementation questions here:

1. do we try to enable BP always in termenv?
   - this seems safe and unlikely to confuse terminals. If a terminal does not support BP, it will ignore the CSI code.
   - however, once we do this we must also ensure that all the bubble widgets know how to handle text pasted this way.

2. how to represent bracketed pastes in bubbletea?
[...]

PR #511, opened 2022-10-08
optimization: use a DFA-based lexer for input

### LINKED ISSUE #502: bug: the input reader gobbles up characters after escape sequences
When a composite input is inserted in the terminal, the terminal can deliver many bytes at once to the process, including escape sequences interleaved with regular keypresses. (This also happens after a process is resumed from sleep and there were input queued while the process was sleeping.)

Currently the input reader only recognizes input boundaries when the read call returns (i.e. no more input to be read).
It is also unable to recognize when an escape sequence is followed by regular bytes (it fails to detect the sequence in the special control character map).

Instead, the input reader should only match a *prefix* to the current input bytes. Then after the prefix is recognized, it should only discard the prefix, then continue to match on the remainder bytes.

This is partly addressed by #397 but it should really be fixed independently of the bracketed paste feature, as it would benefit any composite input.

knz:
First 3 commits from  #568, #569 and #570.

This PR upgrades the map-based sequence detector to a lexical analyzer generated using [golex](https://pkg.go.dev/modernc.org/golex).

The performance improvement is 4x, going down from from ~41.5ns/byte (using a map) to ~18.4ns/byte (using the DFA) on my test computer.

Informs #404.


knz:
cc @muesli for your consideration. This is followup to what we were looking at yesterday. I'll add some benchmarks to showcase the improvement, even in the simple case when there is just 1 event in the input buffer.

knz:
some left over impl notes to self:
- sort sequences to have longest first, to avoid mismatch if we ever have sequences prefix of another in the future.
- randomized testing
- bracketed paste counter-example

knz:
This is ready for review now.

knz:
Note to self: also check the context cancellation during long reads.

knz:
After discussion with @muesli we are considering the following:

- 1 PR to flip the control of readInputs to write the new messages to a channel directly.
- 1 PR to simplify the structure of readInputs while preserving the sequence map
- 1 PR to use the golex-based sequence parser.

knz:
I will need to adapt this after the control flow change in #569.

PR #512, opened 2022-10-08
test: add batchMsg test

muesli:
Simple `batchMsg` test.

### DIFF
--- a/tea_test.go
+++ b/tea_test.go
@@ -7,8 +7,11 @@ import (
 	"time"
 )
 
+type incrementMsg struct{}
+
 type testModel struct {
 	executed atomic.Value
+	counter  atomic.Value
 }
 
 func (m testModel) Init() Cmd {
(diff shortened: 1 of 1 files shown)


tea_test.go
```diff
@@ -84,3 +96,34 @@ func TestTeaKill(t *testing.T) {
...
+	}
+
+	m := &testModel{}
+	p := NewProgram(m, WithInput(&in), WithOutput(&buf))
+	go func() {
+		p.Send(batchMsg{inc, inc})
+
+		for {
+			time.Sleep(time.Millisecond)
+			if m.counter.Load() == 2 {
```
maaslalani, line 115: Though this is _extremely_ unlikely to miss this condition. I usually prefer using `>=` for these types of situations so that we don't accidentally cause a infinitely for loop (albeit _extremely_ unlikely).

```suggestion
			if m.counter.Load() >= 2 {
```

maaslalani: Feel free to ignore this though since it is primarily just paranoia, since it's the likelihood of this happening is negligible.

maaslalani (approved):

tea_test.go
muesli: It's a fair point actually. We could run into this situation if an `incrementMsg` gets duplicated by whatever bug. Luckily tests would still get aborted and never run indefinitely, but it's worth making the change. Thanks for the heads up!

PR #521, opened 2022-10-10
feat: add quit event handler

### LINKED ISSUE #472: Add a way to prevent tea.Quit from terminating the program
*tl;dr: [POC of the proposed feature](https://github.com/aschey/bubbletea/commit/e339dd2864decbf018a67a20b6eac7dfbefa70b5)*

[link to Slack thread where this was originally discussed](https://charmbracelet.slack.com/archives/CQFPW9W87/p1664484606503839)

I'm working on [bubbleprompt](https://github.com/aschey/bubbleprompt), a library very similar to the popular [go-prompt](https://github.com/c-bata/go-prompt). It essentially acts as a REPL where users can run custom actions on submit. In bubbleprompt, I allow consumers of the library to supply an arbitrary `tea.Model` when they submit a command. When this child model sends `tea.Quit`, I need to prevent it from terminating the entire `tea.Program` and instead return control to the parent model.

Currently, I do this using reflection to check if the child model returns `tea.Quit` and prevent it from sending it if so. It looks something like this:
```go
model, cmd := childModel.Update(msg)
if reflect.ValueOf(cmd).Pointer() == reflect.ValueOf(tea.Quit).Pointer() {
    // Don't send the command to the Bubble Tea runtime
    return
{
```
However, if the quit command is wrapped in `tea.Batch` or any similar construct, this gets substantially trickier.

To solve this problem, I propose adding a `tea.onQuit` event handler with a signature such as `onQuit func() QuitBehavior`. If the user supplies it, this function will run whenever the event loop receives a `tea.quitMsg`. The quit handler will return whether the program should shut down or not. 

`QuitBehavior` is just an enum with two values: `tea.Shutdown` and `tea.PreventShutdown`. We could just use a bool instead of a custom enum, but I think that would be rather confusing to the user - does `true` mean shutdown or does `false` mean shutdown? If `onQuit` returns `tea.Shutdown`, then Bubble Tea will handle the message just as it currently does, but if it returns `tea.PreventShutdown`, then we should treat the `tea.quitMsg` as a normal message and forward it to the user's update handler for further processing. This would require making `tea.quitMsg` public so that applications could check for it. 

A minimal usage example would look like this:

```go
type model struct {
    child *tea.Model
}

var shutdown bool = false

func onQuit() tea.QuitBehavior {
    if shutdown {
        return tea.Shutdown
    } else {
        return tea.PreventShutdown
    }
}

    switch msg := msg.(type) {
        case tea.KeyMsg:
            switch msg.Type {
            // If we received a ctrl+c or escape, the user requested the entire program to shut down
            // Set shutdown to true so our quit handler tells the runtime to shut down
            case tea.KeyCtrlC, tea.KeyEsc:
                shutdown = true
                return m, tea.Quit
            }

        // If we received a shutdown message, onQuit returned tea.PreventShutdown which means the 
        // child model initiated the shutdown and our onQuit function prevented the runtime from shutting down.
        // Bubble Tea instead passed the QuitMsg along to our update function so we can handle it however we please.
        case tea.QuitMsg:
            m.child = nil
    }
   
    if m.child != nil {
        return m.child.Update(msg)
    }
    return doSomeOtherAction(msg);
}

func Run(child *tea.Model) {
    m := model{child}
    p := tea.NewProgram(m, tea.WithOnQuit(onQuit))
    p.Start();
}
```

I don't love the use of a global variable here, but I think it's the easiest way to handle this without introducing additional complexity by adding something like another special command that bypasses the `onQuit` handler and always shuts the program down. That feels like a heavy solution for this rather niche use case, so I'm trying to keep the proposed changes to Bubble Tea minimal here.

The implementation is rather simple, so I went ahead and made [a quick POC](https://github.com/aschey/bubbletea/commit/e339dd2864decbf018a67a20b6eac7dfbefa70b5). Happy to make a PR for this after hearing any feedback. Thanks!

aschey:
Resolves #472 

As discussed in #472, this change adds the `WithOnQuit` option so users can intercept quit events and prevent the program from shutting down. This is useful in cases when you want to prompt the user before shutting down or when writing a library on top of Bubble Tea that accepts any `tea.Model` from the user and needs to prevent these child models from shutting the whole program down. This required making `tea.quitMsg` public so users can handle it in their own code if shutdown is prevented. 

I've added an example that demonstrates the former use case (the latter would be hard to model in a simple demo I think). I was hoping to upload a gif to match the other examples, but I couldn't get it to look quite right with the same font size and prompt and everything, so it just looked out of place among the list of other examples. Here's a demo below showing what it looks like.

out

### DIFF
--- a/examples/prevent-quit/main.go
+++ b/examples/prevent-quit/main.go
@@ -1,0 +1,149 @@
+package main
+
+// A program demonstrating how to use the WithOnQuit option to intercept quit events
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/charmbracelet/bubbles/help"
+	"github.com/charmbracelet/bubbles/key"
+	"github.com/charmbracelet/bubbles/textarea"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
+)
+
+var (
+	choiceStyle   = lipgloss.NewStyle().PaddingLeft(1).Foreground(lipgloss.Color("241"))
+	saveTextStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("170"))
+	quitViewStyle = lipgloss.NewStyle().Padding(1).Border(lipgloss.RoundedBorder()).BorderForeground(lipgloss.Color("170"))
+)
+
+func main() {
+	p := tea.NewProgram(initialModel(), tea.WithOnQuit(onQuit))
+
+	if _, err := p.Run(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func onQuit(teaModel tea.Model) tea.QuitBehavior {
+	m := teaModel.(model)
+	if m.hasChanges {
+		return tea.PreventShutdown
+	}
+	return tea.Shutdown
+}
+
+type model struct {
+	textarea   textarea.Model
+	help       help.Model
+	keymap     keymap
+	saveText   string
+	hasChanges bool
+	quitting   bool
+}
+
+type keymap struct {
+	save key.Binding
+	quit key.Binding
+}
+
+func initialModel() model {
+	ti := textarea.New()
+	ti.Placeholder = "Only the best words"
+	ti.Focus()
+
+	return model{
+		textarea: ti,
+		help:     help.NewModel(),
+		keymap: keymap{
+			save: key.NewBinding(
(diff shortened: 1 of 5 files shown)


muesli:
Superseded by #536. Thank you for the high quality PR, @aschey, great work!

PR #536, opened 2022-10-13
feat: add generic event filter

muesli:
`WithFilter` lets you supply an event filter that will be invoked before Bubble Tea processes a `tea.Msg`. The event filter can return any `tea.Msg` which will then get handled by Bubble Tea instead of the original event. If the event filter returns `nil`, the event will be ignored and Bubble Tea will not process it.

As an example, this could be used to prevent a program from shutting down if there are unsaved changes.

Based on the fantastic work by @aschey and supersedes https://github.com/charmbracelet/bubbletea/pull/521.

Might make #114 more relevant.
Resolves https://github.com/charmbracelet/bubbletea/issues/472.

### DIFF
--- a/examples/prevent-quit/main.go
+++ b/examples/prevent-quit/main.go
@@ -1,0 +1,149 @@
+package main
+
+// A program demonstrating how to use the WithOnQuit option to intercept quit events
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/charmbracelet/bubbles/help"
+	"github.com/charmbracelet/bubbles/key"
+	"github.com/charmbracelet/bubbles/textarea"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
+)
+
+var (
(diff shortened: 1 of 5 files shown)


aschey:
Just ported my project to use this and it works great, thanks! I like the more flexible approach.

caarlos0 (approved):
nice!

caarlos0:
@meowgorithm in that case, do you think we would need/want/like to support things like `AND` and `OR` in the filters as well?

muesli:
> @meowgorithm in that case, do you think we would need/want/like to support things like `AND` and `OR` in the filters as well?

Not sure if we need that right away, but this could be done by a proxy struct that wraps several other filters.

meowgorithm:
Boolean operators and/or a proxy could be cool but we should probably merge this feature as-is and address those separately, especially since this feature is (so far) fairly niche.

caarlos0:
yes, that would be my point... we can then wrap lists of filters in structs that implement a filter function or something like that

aschey:
Hi, any chance we could get this merged in? Thanks!

PR #541, opened 2022-10-15
fix: detect terminal size after exec

muesli:
Based on @knz's work in #499, but slightly supersedes this change.

A little more coupling in the resize handling, but a lot less code & logic repetition.

### DIFF
--- a/signals_unix.go
+++ b/signals_unix.go
@@ -4,18 +4,15 @@
 package tea
 
 import (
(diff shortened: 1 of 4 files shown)


knz:
Tip: have you heard about the special `Co-authored-by:` annotation that can be put in commit messages?

knz (approved):

PR #543, opened 2022-10-15
Use latest commit for tutorials

brpratt:
I started going through the tutorial after stumbling upon your [beautiful website](https://charm.sh/). I used `go get ./...` to pull down the `bubbletea` dependency which pulls down version `v0.22.1`, but this release does not include the `.Run()` method (added in https://github.com/charmbracelet/bubbletea/pull/509). Hence, the tutorial in the README includes a compilation error unless you use a non-tagged version.

This PR adds an explicit callout to use `@master`. I thought this might be a nice tip in case others discovery this library before the next tag (assuming it includes the `.Run()` change).

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,12 @@ import (
 )
 ```
 
+Be sure to pull the latest version of the Bubble Tea library for this tutorial.
+
+```
+go get github.com/charmbracelet/bubbletea@master
+```
+
 Bubble Tea programs are comprised of a **model** that describes the application
 state and three simple methods on that model:
 
--- a/tutorials/basics/README.md
+++ b/tutorials/basics/README.md
@@ -31,6 +31,12 @@ import (
 )
 ```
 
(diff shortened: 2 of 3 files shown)


muesli:
The problem with this is that it is incorrect as soon as the README becomes part of a release/archive.

brpratt:
Hi @muesli, thank you for the quick feedback. You make a good point -- using `@master` won't work for source releases.

I can't think of a good way to keep the tutorial in the README compatible with the latest tagged version while also showcasing the latest untagged (and possibly backwards-incompatible) changes. In that case, I suppose this PR is a heads up that anyone who follows the current tutorial (as of cbe309d) while using the latest tagged release (`v0.22.1`) will see a compilation error using `.Run()`. This will no longer be an issue assuming an upcoming `v0.22.2` or `v.0.23.0` release incorporates the changes #509.

PR #549, opened 2022-10-16
docs: minor fixes to comments

winder:
Some misc edits found while inspecting the code.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -47,10 +47,9 @@ type sequenceMsg []Cmd
 // wanted to tick with the system clock every second, minute or hour you
 // could use this. It's also handy for having different things tick in sync.
 //
-// Because we're ticking with the system clock the tick will likely not run for
-// the entire specified duration. For example, if we're ticking for one minute
-// and the clock is at 12:34:20 then the next tick will happen at 12:35:00, 40
-// seconds later.
+// Because we're ticking with the system clock the tick may not run
+// immediately. For example, if we're ticking for one minute and the clock is
+// at 12:34:20 then the next tick will happen at 12:35:00, 40 seconds later.
 //
 // To produce the command, pass a duration and a function which returns
 // a message containing the time at which the tick occurred.
--- a/commands.go
+++ b/commands.go
@@ -47,9 +47,10 @@ type sequenceMsg []Cmd
 // wanted to tick with the system clock every second, minute or hour you
 // could use this. It's also handy for having different things tick in sync.
 //
-// Because we're ticking with the system clock the tick may not run
-// immediately. For example, if we're ticking for one minute and the clock is
-// at 12:34:20 then the next tick will happen at 12:35:00, 40 seconds later.
+// Because we're synchronizing ticks with the system clock, the tick may wait
+// the full duration before running. For example, if we specify one minute and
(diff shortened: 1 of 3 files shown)


commands.go
```diff
@@ -47,10 +47,9 @@ type sequenceMsg []Cmd
 // wanted to tick with the system clock every second, minute or hour you
 // could use this. It's also handy for having different things tick in sync.
 //
-// Because we're ticking with the system clock the tick will likely not run for
-// the entire specified duration. For example, if we're ticking for one minute
-// and the clock is at 12:34:20 then the next tick will happen at 12:35:00, 40
-// seconds later.
+// Because we're ticking with the system clock the tick may not run
+// immediately. For example, if we're ticking for one minute and the clock is
+// at 12:34:20 then the next tick will happen at 12:35:00, 40 seconds later.
```
muesli, line 52: I'm not convinced this change is correct. The original comment is accurate, but I think you may additionally see a slightly delayed start of the command because it'll get enqueued internally first.

mouse.go
```diff
@@ -5,9 +5,10 @@ import (
 	"errors"
 )
 
-// MouseMsg contains information about a mouse event and are sent to a programs
+// MouseMsg contains information about a mouse event and is sent to a programs
```
muesli, line 8: While we're at it, it should read "sent to a program's".

tea.go
```diff
@@ -396,8 +396,7 @@ func (p *Program) Run() (Model, error) {
 		p.renderer = newRenderer(p.output, p.startupOptions.has(withANSICompressor))
 	}
 
-	// Check if output is a TTY before entering raw mode, hiding the cursor and
-	// so on.
+	// Unless the user has specified a custom input, we have ensured that this is a TTY.
```
winder, line 399: I believe this comment has been stale since 0780601791ff7f9b641c0163d1209faf0cc54c16

Regardless, the original tty check (e84314c622c18316016fe2ab164b5f5679895e94) has been gone for a long time.

tea.go
```diff
@@ -116,7 +116,7 @@ type Program struct {
 	ignoreSignals      bool
 
 	// Stores the original reference to stdin for cases where input is not a
-	// TTY on windows and we've automatically opened CONIN$ to receive input.
+	// TTY. On Windows we've automatically opened CONIN$ to receive input.
```
muesli, line 119: This change actually changes the meaning a bit. The original comment is accurate, this is for cases where the input is not a TTY on Windows, and we've automatically opened CONIN$ to receive input.

The comma was probably missing.

tea.go
```diff
@@ -115,8 +115,8 @@ type Program struct {
 	altScreenWasActive bool
 	ignoreSignals      bool
 
-	// Stores the original reference to stdin for cases where input is not a
-	// TTY on windows and we've automatically opened CONIN$ to receive input.
+	// Stores the original reference to stdin on Windows. If the input is not
+	// a TTY, we automatically open CONIN$ to receive input.
```
winder, line 119: I don't think a comma would have been enough to help me. I tried splitting the sentence differently to split up what it's for and when it's used.

I believe my initial confusion was because `where input is not a TTY` sounds like a general statement, so I thought the Windows command was just an example.

PR #550, opened 2022-10-16
chore: clarify handlers type and avoid shadowing

winder:
This variable was shadowing its type. The type name was also very generic, the intent of these handlers was confusing until I finished read through each of the handlers. I believe the new name clarifies things in addition to addressing the shadowing.

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -58,7 +58,7 @@ type Model interface {
 // update function.
 type Cmd func() Msg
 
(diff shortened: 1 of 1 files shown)


bashbunni:
I'm not sure this is necessary

muesli:
I think it might be enough to just add a comment what `handlers` is for, right above it's declaration.

PR #553, opened 2022-10-17
docs: Fix README.md example's main function (`p.Run` -> `p.Start`)

albertito:
There's a bug in the example's `main` function, because `*tea.Program` does not have a `Run` method, it has a `Start` method.

This patch fixes the issue by replacing the `p.Run` call with `p.Start`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,7 +239,7 @@ The last step is to simply run our program. We pass our initial model to
 ```go
 func main() {
     p := tea.NewProgram(initialModel())
-    if _, err := p.Run(); err != nil {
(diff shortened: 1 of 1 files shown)


muesli:
`Run` is a new method in `master` and it sounds like you're using the last stable version of `tea`. I suggest either bumping your `tea` dependency to `master` or switching to the documentation of the last release.

albertito:
Oh, I see, sorry about that and thanks for the explanation!

muesli:
Nothing to apologize for, we always appreciate the effort!

PR #561, opened 2022-10-20
feat(ci): auto go mod tidy examples

caarlos0:
this should prevent having to manually run `go mod tidy` in the `examples` folder.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -1,0 +1,34 @@
+name: examples
+
+on:
+  push:
+    branches:
+      - 'master'
+    paths:
+      - '.github/workflows/examples.yml'
+      - './examples/go.mod'
+      - './examples/go.sum'
+      - './go.mod'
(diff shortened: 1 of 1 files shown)


caarlos0:
cc/ @aymanbagabas

.github/workflows/examples.yml
```diff
@@ -0,0 +1,38 @@
...
+    branches:
+      - 'master'
+    paths:
+      - '.github/workflows/examples.yml'
+      - './examples/go.mod'
+      - './examples/go.sum'
+      - './tutorials/go.mod'
+      - './tutorials/go.sum'
+      - './go.mod'
+      - './go.sum'
```
aymanbagabas, line 14: Is this intended?

caarlos0: yep, if we change something in the root module, since the other two replace it, we should tidy them as well

aymanbagabas (approved):

PR #567, opened 2022-10-21
Support changing background color

### LINKED ISSUE #207: Set terminal background color
Hey, thanks for the fantastic set of libraries!

It would be great to have a way to set the background color for the entire terminal, which would be reset upon exit. This is useful for apps that use AltScreen.

michal-84a:
Closes #207 

This PR adds support for changing background color of a terminal and resets it back to the original color when a program exits.
The color can be changed by sending a special internal message, for example in the init function, such as:

```go
  return tea.Cmd(func() tea.Msg {
    return tea.SetBackgroundColor(termenv.RGBColor("#eeeeee"))
  })
}
```

This is the first method in the `renderer` which accepts an argument.
Happy to discuss alternatives in case this breaks renderer's design.

The method `SetBackgroundColor(c termenv.Color) Msg` could be changed to accept hex colors via a string `SetBackgroundColor(c string) Msg`, but the former provides more flexibility with color definition format.

### DIFF
--- a/nil_renderer.go
+++ b/nil_renderer.go
@@ -1,19 +1,23 @@
 package tea
 
+import "github.com/muesli/termenv"
+
 type nilRenderer struct{}
 
-func (n nilRenderer) start()                  {}
-func (n nilRenderer) stop()                   {}
-func (n nilRenderer) kill()                   {}
-func (n nilRenderer) write(v string)          {}
-func (n nilRenderer) repaint()                {}
-func (n nilRenderer) clearScreen()            {}
-func (n nilRenderer) altScreen() bool         { return false }
-func (n nilRenderer) enterAltScreen()         {}
-func (n nilRenderer) exitAltScreen()          {}
-func (n nilRenderer) showCursor()             {}
-func (n nilRenderer) hideCursor()             {}
-func (n nilRenderer) enableMouseCellMotion()  {}
-func (n nilRenderer) disableMouseCellMotion() {}
-func (n nilRenderer) enableMouseAllMotion()   {}
-func (n nilRenderer) disableMouseAllMotion()  {}
+func (n nilRenderer) start()                             {}
+func (n nilRenderer) stop()                              {}
+func (n nilRenderer) kill()                              {}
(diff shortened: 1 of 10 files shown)


muesli:
Nice work, just tested this and it seems to work just fine. Code looks pretty good, too. A few things I would change:

1. Retrieving the current background color of a terminal works in most circumstances, but not always. We should only reset the background color on shutdown if it has actually been manually set before.
2. We should add a test (similar to the other tests in `screen_test.go`).
3. We should probably add an example before we merge this.

I'm wondering if this shouldn't (also) be a `ProgramOption` however. Similar to `AltScreen` it's something you're probably(?) not going to change while your program is running.

michal-kralik:
Thank you for your feedback.
I agree, I'll submit the updates.

I considered the `ProgramOption` but figured if you only had the `ProgramOption` available, you would not be able to change the background later. You may want to change it to red in case of an "catastrophic event" for example :)

Happy to implement the `ProgramOption` as an alternative though. It shall be easier to understand for the users how to use it.

muesli:
> Happy to implement the `ProgramOption` as an alternative though. It shall be easier to understand for the users how to use it.

Oh yeah sorry, I didn't mean to suggest to remove the command entirely, adding the alternative `ProgramOption` would be great though.

michal-kralik:
@muesli this is ready for review now

tcurdt:
would love to see this merged.

michal-kralik:
@muesli feel free to close this PR if you don't feel it adds value.

muesli:
> @muesli feel free to close this PR if you don't feel it adds value.

Definitely adds value and it's something we want! Will come back to reviewing this in a bit!

coxley:
Would love this as well. :)

djyuhn:
Hey! Would love to have this merged if all looks good.

shalldie:
Any progress ?

aymanbagabas:
Fwiw, this now implemented in Bubble Tea v2. You can try it here https://github.com/charmbracelet/bubbletea/releases/tag/v2.0.0-alpha.1

PR #568, opened 2022-10-21
fix(key),test: simplify the input analysis code

knz:
This PR contains only the input simplification part of #511.

cc @muesli

### DIFF
--- a/key.go
+++ b/key.go
@@ -1,8 +1,9 @@
 package tea
 
 import (
-	"errors"
+	"fmt"
 	"io"
+	"regexp"
 	"unicode/utf8"
 
 	"github.com/mattn/go-localereader"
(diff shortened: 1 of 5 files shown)


key.go
```diff
@@ -336,87 +337,77 @@ var keyNames = map[KeyType]string{
 }
 
 // Sequence mappings.
+//
+// Note: if you modify this map, do not forget to re-run `go generate`.
```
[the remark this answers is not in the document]
knz: Good point. Removed.

knz:
@muesli this is ready again.

key_test.go
```diff
@@ -191,34 +384,236 @@ func TestReadInput(t *testing.T) {
...
 			}
+		})
+	}
+}
 
-			for i, v := range msgs {
-				if m, ok := v.(KeyMsg); ok &&
-					m.String() != td.out[i].(KeyMsg).String() {
-					t.Fatalf(`expected a keymsg %q, got %q`, td.out[i].(KeyMsg), m)
+// randTest defines the test input and expected output for a sequence
```
muesli, line 478: Good enough for now, but I guess we could eventually replace the randomizer with a proper fuzzer?

knz: Yes that is reasonable. Do you want to do this right away? Is there something standard we can use?

muesli: Let's keep it as it is, the code looks and works fine for me. As soon as we bump Go to 1.18 we can start moving to the Go fuzzer that's part of the standard toolchain: https://go.dev/security/fuzz/

PR #569, opened 2022-10-21
fix(key): invert the control loop

knz:
First commit from #568.

This is the control flow change discussed in https://github.com/charmbracelet/bubbletea/pull/511#issuecomment-1287061948

cc @muesli


knz:
@muesli this is ready again.

PR #570, opened 2022-10-21
fix(key): support very long buffered input

### LINKED ISSUE #502: bug: the input reader gobbles up characters after escape sequences

knz:
Fixes #502.

First two commits from #568 and #569. 

Split away from #511 as discussed in https://github.com/charmbracelet/bubbletea/pull/511#issuecomment-1287061948

cc @muesli

### DIFF
--- a/key.go
+++ b/key.go
@@ -644,9 +644,9 @@ func detectOneMsg(b []byte, canHaveMoreData bool) (w int, msg Msg) {
 	if i >= len(b) && canHaveMoreData {
 		// We have encountered the end of the input buffer. Alas, we can't
 		// be sure whether the data in the remainder of the buffer is
-		// complete (maybe there was a short read). Instead of saying anything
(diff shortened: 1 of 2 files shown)


knz:
@muesli this is ready again.

Issue #564, opened 2022-10-20
README Documentation on master is ahead of latest release for Run()


taigrr:
Reading through the README, Commit #6b77c8f changes the usage of Run() vs Start() vs StartReturningModel() 12 days ago, but the most recent release was Aug 19. Therefore, the instructions provided can't be followed without using a go replace directive or a workspace file.

I would request this issue stays open for those new to bubbletea as a helpful pointer until a new tagged release becomes available.

Pointer: until a release after v0.22.1 becomes available, use `StartReturningModel()` in place of `Run()`

enrichman:
Thanks, I've encountered the same issue.

taigrr:
New code paths are available after point release! Closing.

PR #584, opened 2022-10-30
Update p.Run() to p.Start()

cmstoddard:
After copy pasting the code to start toying with it, I got an error: p.Run undefined (type *tea.Program has no field or method Run)compilerMissingFieldOrMethod. P.Start() doesn't error, and seems like the correct method. My GO version: go version go1.19.1 darwin/arm64. Using v0.22.1


muesli:
See https://github.com/charmbracelet/bubbletea/issues/564

PR #590, opened 2022-11-08
[BUG] Update main.go

zarifpour:
Currently, the file does not run because of the error:
```
p.Run undefined (type *tea.Program has no field or method Run)
```
`Run` can be replaced with the method `StartReturningModel` to resolve the error.


muesli:
See #589 and others. You're looking at the docs of the `master` branch while trying to build against the last stable release. `Run` is a new method that replaces `StartReturningModel`.

Issue #359, opened 2022-07-01
bug: windows missing mouse events past column 95


lrstanley:
Running the [mouse](https://github.com/charmbracelet/bubbletea/blob/master/examples/mouse/main.go) example on powershell, Windows Terminal, cygwin-style terminals, and potentially others, doesn't seem to let me capture mouse events beyond column 95. I haven't tested on a proper linux environment yet

Example video: https://ls-screen.s3.us-west-004.backblazeb2.com/2022/07/WindowsTerminal_ninSH9gbiM.mp4

Looking into this further, I did find [this issue comment](https://github.com/microsoft/terminal/issues/545#issuecomment-509436962), which mentions the following:
> The legacy, byte-based protocol only allows row and column numbers up to 223, because 32 is added to this number and this is sent as a single byte. The column limit is not uncommon to be too small. (By the way, **beginning at column 95 the generated data is not 7-bit clean and not valid UTF-8** which is a problem for encoding conversion layers such as luit.)

Specifically:
> beginning at column 95 the generated data is not 7-bit clean and not valid UTF-8

That issue was closed and features implemented, mentioning `SGR 1006 extension`, however I'm not sure if it's something on my end, something on Windows end, or an issue with bubbletea. 😅 

This SO post is unrelated to bubbletea/windows terminal, but also seems to explain the mouse events (legacy vs 1006) to me a little bit more: https://stackoverflow.com/a/5970472/1830159

Any pointers?

lrstanley:
Actually, I don't know why I didn't think to test another application, but it looks like Vim also has a similar issue when enabling mouse mode, so I don't think it's a bubbletea issue. Though I'm still curious if anyone has any pointers (ha!) on what the issue could be. I've got the latest Windows Terminal version, and the latest Windows 10 update, and I can't seem to find others with related issues when searching, so I'm at a loss.

lrstanley:
Just in case anyone else runs into similar issues, upgraded to Windows 11 recently, and the issues disappeared, so I'm guessing something related to WSL2 and Windows 10. Going to close out.

muesli:
Oh wow, that's unexpected and great to hear! Thanks for the feedback!

DerpyDashie:
#594 should fix this

PR #594, opened 2022-11-10
Extended Coordinates mouse reporting & additional buttons support

aymanbagabas:
The X10 mouse protocol limits the X and Y coordinates to 223. This adds SGR (1006) mode to enable extended mouse events support.

* Add <kbd>shift</kbd> modifier support
* Support left/right wheel buttons (codes 6 & 7)
* Support additional mouse button (codes 8-11)
* Report which button was released (SGR feature)

Wants: https://github.com/muesli/termenv/pull/104

Fixes: https://github.com/charmbracelet/bubbletea/issues/359
Fixes: https://github.com/charmbracelet/bubbletea/issues/718


examples/mouse/main.go
```diff
@@ -4,21 +4,19 @@ package main
...
 import (
-	"fmt"
 	"log"
 
 	tea "github.com/charmbracelet/bubbletea"
 )
 
 func main() {
-	p := tea.NewProgram(model{}, tea.WithAltScreen(), tea.WithMouseAllMotion())
+	p := tea.NewProgram(model{}, tea.WithMouseAllMotion(), tea.WithMouseExtendedMode())
```
meowgorithm, line 13: Do we need both of these? Ideally, would be nice to get all the benefits of SGR mode with just one flag.

mouse.go
```diff
@@ -28,38 +46,216 @@ func (m MouseEvent) String() (s string) {
...
+var mouseButtons = map[MouseButton]string{
+	MouseButtonNone:       "none",
+	MouseButtonLeft:       "left",
+	MouseButtonMiddle:     "middle",
+	MouseButtonRight:      "right",
+	MouseButtonWheelUp:    "wheel up",
+	MouseButtonWheelDown:  "wheel down",
+	MouseButtonWheelLeft:  "wheel left",
+	MouseButtonWheelRight: "wheel right",
+	MouseButtonBackward:   "backward",
```
meowgorithm, line 130: I'd probably rename this (and the deprecated version) to `MouseButtonBack` so it's super clear that it's the back button and not a backwards motion.

mouse.go
```diff
@@ -28,38 +46,216 @@ func (m MouseEvent) String() (s string) {
...
+	MouseMiddle:     "middle",
+	MouseRelease:    "release",
+	MouseWheelUp:    "wheel up",
+	MouseWheelDown:  "wheel down",
+	MouseWheelLeft:  "wheel left",
+	MouseWheelRight: "wheel right",
+	MouseBackward:   "backward",
+	MouseForward:    "forward",
+	MouseBtn10:      "button 10",
+	MouseBtn11:      "button 11",
```
meowgorithm, line 175: I almost wonder if we need to add these two at all since we're deprecating this API anyway.

aymanbagabas: Good call!

examples/mouse/main.go
aymanbagabas: We could have SGR mode enabled by default when using `WithMouseCellMotion` and `WithMouseAllMotion`, and then have `WithMousePixelsMode` to use SGR-Pixels.

robin-samuel:
You have to fix the mouseEventLen in key.go otherwise it will throw after each detected sgr event random keyRunes Events with left over chars. Bad for focused input fields and so on. SGR events dont have like X10 Events a fixed length of 6. Otherwise very cool works very clean on windows!

My recommended solution:

```go
func detectOneMsg(b []byte) (w int, msg Msg) {
	// Detect mouse events.
	// X10 mouse events have a length of 6 bytes
	const mouseEventX10Len = 6
	if len(b) >= mouseEventX10Len && b[0] == '\x1b' && b[1] == '[' {
		switch b[2] {
		case 'M':
			return mouseEventX10Len, MouseMsg(parseX10MouseEvent(b))
		case '<':
			if matchIndices := mouseSGRRegex.FindSubmatchIndex(b[2:]); matchIndices != nil {
				// SGR mouse events length is the length of the match plus the length of the escape sequence
				mouseEventSGRLen := matchIndices[1] + 2
				return mouseEventSGRLen, MouseMsg(parseSGRMouseEvent(b))
			}
		}
	}
	...
}
 
```

Would love to see this soon be merged into main branch. 😄

key.go
```diff
@@ -570,11 +570,21 @@ func readInputs(ctx context.Context, msgs chan<- Msg, input io.Reader) error {
...
+	// X10 mouse events have a length of 6 bytes
 	const mouseEventLen = 6
-	if len(b) >= mouseEventLen && b[0] == '\x1b' && b[1] == '[' && b[2] == 'M' {
-		return mouseEventLen, MouseMsg(parseX10MouseEvent(b))
+	if len(b) >= mouseEventLen && b[0] == '\x1b' && b[1] == '[' {
+		switch b[2] {
+		case 'M':
+			return mouseEventLen, MouseMsg(parseX10MouseEvent(b))
+		case '<':
+			if mouseSGRRegex.Match(b[3:]) {
```
robin-samuel, line 584: You have to fix the mouseEventLen in here otherwise it will throw after each detected sgr event random keyRunes Events with left over chars. Bad for focused input fields and so on. SGR events dont have like X10 Events a fixed length of 6. Otherwise very cool! Works very clean on windows!

My recommended solution:

```go
func detectOneMsg(b []byte) (w int, msg Msg) {
	// Detect mouse events.
	// X10 mouse events have a length of 6 bytes
	const mouseEventX10Len = 6
	if len(b) >= mouseEventX10Len && b[0] == '\x1b' && b[1] == '[' {
		switch b[2] {
		case 'M':
			return mouseEventX10Len, MouseMsg(parseX10MouseEvent(b))
		case '<':
			if matchIndices := mouseSGRRegex.FindSubmatchIndex(b[3:]); matchIndices != nil {
				// SGR mouse events length is the length of the match plus the length of the escape sequence
				mouseEventSGRLen := matchIndices[1] + 3
				return mouseEventSGRLen, MouseMsg(parseSGRMouseEvent(b))
			}
		}
	}
	...
}
 
```

Would love to see this soon be merged into main branch. 😄

robin-samuel (changes requested):

key.go
```diff
@@ -570,11 +570,21 @@ func readInputs(ctx context.Context, msgs chan<- Msg, input io.Reader) error {
...
+				return mouseEventLen, MouseMsg(parseSGRMouseEvent(b))
+			}
+		}
 	}
```
robin-samuel, line 588: You have to fix the mouseEventLen here otherwise it will throw after each detected sgr event random keyRunes Events with left over chars. Bad for focused input fields and so on. SGR events dont have like X10 Events a fixed length of 6. Otherwise very cool. Works very clean on windows!

My recommended solution:

```suggestion
func detectOneMsg(b []byte) (w int, msg Msg) {
	// Detect mouse events.
	// X10 mouse events have a length of 6 bytes
	const mouseEventX10Len = 6
	if len(b) >= mouseEventX10Len && b[0] == '\x1b' && b[1] == '[' {
		switch b[2] {
		case 'M':
			return mouseEventX10Len, MouseMsg(parseX10MouseEvent(b))
		case '<':
			if matchIndices := mouseSGRRegex.FindSubmatchIndex(b[3:]); matchIndices != nil {
				// SGR mouse events length is the length of the match plus the length of the escape sequence
				mouseEventSGRLen := matchIndices[1] + 3
				return mouseEventSGRLen, MouseMsg(parseSGRMouseEvent(b))
			}
		}
	}
 
```

Would love to see this soon be merged into main branch. 😄

robin-samuel (changes requested):

key.go
aymanbagabas: @robin-samuel good catch! It would be nice to have a test case for this 🙂

robin-samuel:
```go
package main

import (
	"github.com/charmbracelet/bubbles/textinput"
	"github.com/charmbracelet/lipgloss"
)

type Model struct {
	width  int
	height int
	input  textinput.Model
}

func New() *Model {
	ti := textinput.New()
	ti.Placeholder = "Type something..."
	ti.Focus()
	return &Model{
		input: ti,
	}
}

func (m *Model) Init() tea.Cmd {
	return textinput.Blink
}

func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	cmds := make([]tea.Cmd, 0)
	switch msg := msg.(type) {
	case tea.WindowSizeMsg:
		m.width = msg.Width
		m.height = msg.Height
	case tea.KeyMsg:
		switch msg.Type {
		case tea.KeyCtrlC:
			return m, tea.Quit
		}
	}
	ti, cmd := m.input.Update(msg)
	m.input = ti
	if cmd != nil {
		cmds = append(cmds, cmd)
	}
	return m, tea.Batch(cmds...)
}

func (m *Model) View() string {
	style := lipgloss.NewStyle().
		Width(m.width).
		Height(m.height)
	return style.Render(lipgloss.JoinVertical(lipgloss.Left,
		"Move your mouse around over the terminal window.",
		"You will see random inputs appear in the text input field below.",
		m.input.View(),
		"Press Ctrl+C to quit.",
	))
}

func main() {
	p := tea.NewProgram(New(), tea.WithAltScreen(), tea.WithMouseAllMotion())
		panic(err)
	}
}
```

> @robin-samuel good catch! It would be nice to have a test case for this 🙂

Here a very simple example. By moving the mous over the window you will see that the input gets spammed with key.Runes events and this text will automatically be put in the input field. With the fix applied this wont be the case and act like usual! 🙂

aymanbagabas: @robin-samuel thank you 🙂

meowgorithm (approved):

PR #597, opened 2022-11-13
docs: update delve information

aschey:
I ran into a bit of difficulty trying to connect the VS Code debugger to the process spawned by `dlv debug`. I eventually figured out that delve defaults to its legacy protocol (version 1) by default which was causing compatibility issues with VS Code's Go integration. Delve's docs recommend using version 2 for all new development, so I think it makes sense to add that to the docs so people trying to integrate with IDEs hopefully don't run into the same compatibility issues. Also added a note about supplying the port so these commands can integrate with IDEs and shell scripts without having to figure out the port each time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,15 +266,21 @@ delve in headless mode and then connect to it:
 
 ```bash
 # Start the debugger
-$ dlv debug --headless .
-API server listening at: 127.0.0.1:34241
+$ dlv debug --headless --api-version=2 --listen=127.0.0.1:43000 .
+API server listening at: 127.0.0.1:43000
 
 # Connect to it from another terminal
-$ dlv connect 127.0.0.1:34241
+$ dlv connect 127.0.0.1:43000
 ```
 
-Note that the default port used will vary on your system and per run, so
-actually watch out what address the first `dlv` run tells you to connect to.
+If you do not explicitly supply the `--listen` flag, the port used will vary
+per run, so passing this in makes the debugger easier to use from a script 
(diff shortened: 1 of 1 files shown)


meowgorithm:
@muesli does this seem correct to you (and is this still relevant @aschey)? I'm skimming the delve docs and don't see anything on this, so I wonder if it's the default now.

aschey:
I believe this is still relevant. The specific documentation I remember reading is this: https://github.com/go-delve/delve/blob/master/Documentation/api/ClientHowto.md#selecting-the-api-version

> Delve currently supports two version of its API, APIv1 and APIv2. By default a headless instance of dlv will serve APIv1 for backward-compatibility with older clients, however new clients should use APIv2 as new features will only be made available through version 2. The preferred method of switching to APIv2 is to send the RPCServer.SetApiVersion command right after connecting to the backend. Alternatively the --api-version=2 command line option can be used when spawning the backend.

meowgorithm (approved):

maaslalani:
@meowgorithm Is this good to merge?

meowgorithm:
Sorry to take so long as this. @maaslalani, can we verify this is still current?

maaslalani:
Yes, this is still current from my understanding. The `dlv` CLI mentions that the default `--api-version` is `1` but we should use `2`.

image

meowgorithm:
Done and done! Thanks for the contribution here @aschey and sorry it took forever!

Issue #602, opened 2022-11-16
feat: guarantee that the first message every program always receives is the tea.WindowSizeMsg


bashbunni:
We often see a problem where models are not created until a `tea.WindowSizeMsg` is received so it can load in with the correct sizes. However, this creates a common race condition where something tries to set the values of this bubble/model before it has been initialized as another message is received before the `WindowSizeMsg`, leading to a nil pointer error.

In the meantime having this issue and maybe a disclaimer in the README would be helpful for users trying to debug the issue. 

Real world example of this issue: https://github.com/orlangure/gocovsh/issues/38

deicon:
If *tea.WindowSizeMsg* is really and in all cases the indication that the application had a chance to properly init all models, all other messages could be defered/buffered until the *tea.WindowSizeMsg* is pushed on the msgs channel.

Basically by waiting on a mutex which gets released once the *tea.WindowSizeMsg* is pushed into Update()

meowgorithm:
This can and should be solved in documentation.

On the technical side of things, the initial view is called before the first update which means that any view logic around unknown terminal size will need to be accounted for anyway.

More importantly, however, Bubble Tea programs are asynchronous in nature and, as a rule, message order is not guaranteed. If we encourage developers to design their programs in such a manner they'll end up with much more robust applications overall.

If we start to guarantee message order on startup we also risk complicating the mental model. What order do initial messages come in? Do they happen before or after `Init()`? And so on.

In the case of `gocovish`, the solve is to guard against the nil pointer error in [`list`](https://github.com/charmbracelet/bubbles#list).

PR #605, opened 2022-11-16
Feature/602/assure window size msg comes first

deicon:
#602 Naive approach using a Program wide mutex to block all but the WindowSizeMsg until the WindowSizeMsg was received and processed in the Update method. Once this is done, the Program is marked with initDone and the mutex is no longer blocking the custom messages. 

Not sure about this approach. Maybe a reviewer has a look at it

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -123,6 +123,7 @@ type Program struct {
 	// below.
 	windowsStdin *os.File //nolint:golint,structcheck,unused
 	initMtx      sync.Mutex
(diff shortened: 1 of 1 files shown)


meowgorithm:
Hey @deicon! Thank you for the PR. For now I'm going to recommend we hold off on this per my comment in https://github.com/charmbracelet/bubbletea/issues/602. We can reexamine this down the road if things change.

PR #624, opened 2022-12-16
docs: add brows to "in the wild" app list

rubysolo:


maaslalani:
Very cool app @rubysolo, thanks so much for sharing it!!!

PR #627, opened 2022-12-24
fix: replaced the old version function with new version used p.Start(

Prajwalprakash3722:
we are ignoring the model returned by the NewProgram can also use p.StartReturningModel() if we plan to use the model in the code.


muesli:
`Start` is deprecated and will be removed in a future release. As such I think we should stick with `Run` here.

Prajwalprakash3722:
I apologise for my ignorance, I'm new to this codebase, still getting the hang of it.

PR #633, opened 2023-01-04
docs: add SlurmCommander to in-the-wild section

pja237:


meowgorithm:
_Slurm_ is such a great word. Thanks for the PR!

PR #636, opened 2023-01-15
feat(frame-rate): add configurable frame-rate option

### LINKED ISSUE #577: Feature request: Configure the max frame rate
The default Framerate is defined here:
https://github.com/charmbracelet/bubbletea/blob/2ebe7426db048be059cdc56a0e0437266bf9706b/standard_renderer.go#L17

It is set for 60 fps, which for some cases, where network lag exists, is too much.
I want to be able to configure the `standardRenderer` max FPS but currently there is no option to do so

TheYahya:
By this changes, we will be able to set frame rate before starting our program.

```go
p := tea.NewProgram(&m, tea.WithAltScreen())
p.FrameRate(time.Second) // setting frame rate to one second
if err := p.Start(); err != nil {
        fmt.Println("Error running program:", err)
        os.Exit(1)
}
```

fix #577

### DIFF
--- a/nil_renderer.go
+++ b/nil_renderer.go
@@ -1,5 +1,7 @@
 package tea
 
+import "time"
+
 type nilRenderer struct{}
 
 func (n nilRenderer) start()                  {}
@@ -17,3 +19,4 @@ func (n nilRenderer) enableMouseCellMotion()  {}
 func (n nilRenderer) disableMouseCellMotion() {}
 func (n nilRenderer) enableMouseAllMotion()   {}
 func (n nilRenderer) disableMouseAllMotion()  {}
+func (n nilRenderer) frameRate(time.Duration) {}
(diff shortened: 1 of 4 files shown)


meowgorithm:
Related #636

meowgorithm:
Hi! This doesn't actually solve #636 because it sets a constant framerate as opposed to a max frame rate.

Regardless, if we decide this is something we want to go with with PR will need to:

1. Introduce a `ProgramOption` for this, i.e. `tea.WithFramerate(30)`
1. Change variable naming should drop camelcasing to match the variable under the hood: `FrameRate` -> `Framerate`
1. Decide if it makes sense for one to be able to change the framerate after the program has started. If so, we'll also want to introduce a command for that so it can be changed from within a program. If not, `Program.Framerate` should not be exposed.

Issue #639, opened 2023-01-20
`eventLoop` `sequenceMsg` should check `cmd` is not `nil`


nderjung:
Supplying a cmd which is `nil` (which may be mixed in a sequence of `cmds`) will result in a nil pointer exception when the [`eventLoop` iterates over the list of `msg`](https://github.com/charmbracelet/bubbletea/blob/master/tea.go#L313).

This issue tracks a simple `if cmd != nil` to be added before invoking `cmd()`.

meowgorithm:
Thanks so much for catching this one! Merging your PR shortly.

PR #640, opened 2023-01-20
fix: Check msg cmd is not nil before invoking

nderjung:
GitHub-Fixes: #639


meowgorithm:
Good catch—this makes sense to me. @muesli, look good to you?

muesli:
Looks good to me, too. We do a similar check for batched messages.

Issue #654, opened 2023-02-08
Can't figure out how to show error in dedicated field in credit-card-example


gattiroberto1985:
Hello,

i see that in the `credit-card-form` example [here](https://github.com/charmbracelet/bubbletea/blob/master/examples/credit-card-form/main.go) that an `error` is stored in the model and an `errMsg` message is handled in the `Update` function. But i cannot make the error visible in the UI, nothing appears if i input a wrong character, neither if i print the error explicitly in the `View` function (if the error is not `nil`). It seems that the `errMsg` is never activated. 

I can't find any documentation or example about this, is there something wrong in the example?
Thanks in advance!

hopefulTex:
The validation functions are called by the text models whenever one receives the `Update(teaMsg)` call, all you have to do is pass it a message and the rest is taken care of.  As for the `errMsg` , that's going to have whatever errors are passed into your program model itself, while the text inputs do error handling in their own `Update()` methods.
The errors are stored in the individual textinputs, `m.inputs[n].Err` , and they're set to `nil` if there's no error.

bashbunni:
Hey @hopefulTex,

This is an issue that's being [discussed](https://github.com/charmbracelet/bubbles/pull/185) upstream in Bubbles. It would make more sense to continue the discussions and apply improvements over there. Thank you for bringing this to our attention! We haven't forgotten about this issue

PR #660, opened 2023-02-14
feat(examples): Credit Card Example Now Displays User Input Errors

hopefulTex:
At the request of, and as a response to #654 ,
While the credit card example may have shown developers how to use the `textinput` bubble, the application itself was lacking in visual feedback, both for the "user" as well as the developer.
This change, gives a more natural and responsive UI for developers as they iterate on the example to further learn the library.
  The first two commits are nullified, I can clean if yall'd like.

### DIFF
--- a/examples/credit-card-form/main.go
+++ b/examples/credit-card-form/main.go
@@ -47,6 +47,7 @@ type model struct {
 }
 
 // Validator functions to ensure valid input
+// Note: Please don't mix runtime errors with your users' errors
 func ccnValidator(s string) error {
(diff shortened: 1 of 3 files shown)


bashbunni:
This is a fix that needs to happen in the upstream [Bubbles repo](https://github.com/charmbracelet/bubbles). We are doing the validation there for the input fields and there is an open PR regarding this issue: https://github.com/charmbracelet/bubbles/pull/185

Sorry for any confusion this may have caused with the credit card example. Thank you for your support!!

PR #667, opened 2023-02-16
Add WG Commander to README.md

AndrianBdn:
I have recently created an open-source tool for simple WireGuard configuration management. 

bubbletea (as well as lipgloss) is a delight to use. I would be happy if it is possible to mention my project on the bubbletea page. 

Thank you.


meowgorithm:
This is such a fun one. Thanks for adding!

PR #673, opened 2023-02-19
fix: update x/net

caarlos0:
CVE-2022-27664


caarlos0:
already fixed on master

PR #720, opened 2023-04-12
fix: do not create an output from an output

caarlos0:
if the user uses `WithOutput(o)`, and `o` is already a `*termenv.Output`, hell breaks loos (at least if you do it over SSH, haven't tested plain).

Anyway, we should probably not create a new output from an existing output? if so, this should do it.


aymanbagabas:
@caarlos0 this is a dup of https://github.com/charmbracelet/bubbletea/pull/715 :)

cc/ @muesli @meowgorithm

caarlos0:
oh, missed it!

closing this one then

PR #722, opened 2023-04-17
feat: tea.Wait

caarlos0:
wait for the underlying context to finish.

extract from #352

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -97,8 +97,9 @@ type Program struct {
 	ctx    context.Context
 	cancel context.CancelFunc
 
-	msgs chan Msg
-	errs chan error
+	msgs     chan Msg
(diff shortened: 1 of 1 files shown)


aymanbagabas (approved):

meowgorithm (commented):
Per my chat with @muesli, it would be awesome if `Wait()` could return after the terminal state has been restored (internally, at the end of `Program.shutdown`).

caarlos0:
ah, I actually don't even need this, I can pass my own context with `WithContext`.

as per your request, pushed something I think will help

meowgorithm (approved):
Nice! This is exactly how I solved for it in #726 — except your API design's much better.

PR #723, opened 2023-04-17
feat: IsQuitCmd & IsQuitMsg

caarlos0:
Allows to check if the given command/message will cause the application
to quit.

Extracted from #352

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -134,6 +134,26 @@ func Quit() Msg {
 // send a quitMsg with Quit.
 type quitMsg struct{}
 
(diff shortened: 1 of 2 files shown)


muesli:
Heads up: I'm still leaning towards making `QuitMsg` public: https://github.com/charmbracelet/bubbletea/pull/536/files

Might be the nicer solution instead of adding public API we'll eventually deprecate again.

meowgorithm:
Considering `BatchMsg` is public now I don't see why `QuitMsg` shouldn't be public. In general I think #536 is fine to be merged at this point.

PR #726, opened 2023-04-18
feat: add method for blocking until shutdown is complete

meowgorithm:
This feature introduces a `Program.Done()` method which will block until a `Program` has completely exited. This is useful when you want to wait until Bubble Tea has completely restored the terminal before printing output elsewhere in your Go application.

In particular, this feature can help prevent race conditions where Bubble Tea and other output are attempting to print to the terminal at the same time.

Consider the following:

```go
// Draw a spinner
go func() {
		fmt.Println("Uh oh:", err)
		return
	}
}()

result := someLongRunningOperation()

p.Quit()   // stop the spinner
<-p.Done() // wait for cleanup

// Print safely.
fmt.Println("Okay, so the results are in:", result)
```

It may also make sense to include a convenience function which combines `Program.Quit` and `Program.Done`. Potentially named `Program.Shutdown()`?

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -115,6 +115,10 @@ type Program struct {
 	altScreenWasActive bool
 	ignoreSignals      bool
 
+	// Used to signal that the Program has completely finished running. This
+	// a
+	shutdownComplete chan struct{}
+
 	// Stores the original reference to stdin for cases where input is not a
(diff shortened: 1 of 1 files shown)


meowgorithm:
Closing in favor of #722, which has a much nicer API!

PR #728, opened 2023-04-22
docs(readme): Add ascii-movie to 'Bubble Tea in the Wild' section

gabe565:
Hi! I've used Bubble Tea in a few different projects, but finally have one that feels right to add to the list of example programs. This PR adds a link to [gabe565/ascii-movie](https://github.com/gabe565/ascii-movie) to the readme.

ASCII Movie plays the Star Wars telnet movie locally, over SSH, or over Telnet and uses Bubble Tea for its TUI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,6 +316,7 @@ your program in another window.
 
 For some Bubble Tea programs in production, see:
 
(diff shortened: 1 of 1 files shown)


meowgorithm:
Thanks for the PR!

PR #729, opened 2023-04-22
fix(examples): print all packages in package manager

Madh93:
This PR include the next small corrections related with the [package manager example](https://github.com/charmbracelet/bubbletea/tree/master/examples/package-manager):

- The packages counter in the progress bar is incorrect
- The latest package is not printed

**Before**

before

**After**

after


bubylou (approved):
Noticed this issue as well and confirmed this fixes it.

bubylou (approved):

meowgorithm:
Thanks for bumping this, @bubylou, and thanks so much for the fix here @Madh93! We meant to merge this one right when it came in.

@Madh93, do you mind rebasing this one on `main`? Once that's done we'll merge it in.

PR #733, opened 2023-05-05
feat: WithoutInput ProgramOption for disabling input entirely

meowgorithm:


muesli:
Wouldn't `WithInput(nil)` suffice here?

meowgorithm:
Closed because `WithInput(nil)` indeed suffices :joy:

PR #734, opened 2023-05-05
chore: make input options mutually exclusive

meowgorithm:
Just some internal housekeeping: basically `WithCustomInput` and `WithInputTTY` can not longer both exist in `startupOptions` internally. When both are passed as `ProgramOption`s the last one wins.


muesli (approved):
LGTM!

meowgorithm:
Thanks @muesli!

PR #735, opened 2023-05-08
chore: update example tests, test on ci

caarlos0:
- run tests on examples on CI too
- update teatest


.github/workflows/build.yml
```diff
@@ -32,6 +32,10 @@ jobs:
         run: go build -v ./...
         working-directory: ./examples
 
+      - name: Test examples
```
bashbunni, line 35: Should we at some point have our workflows here pointing to `meta`? Like we do in Charm (https://github.com/charmbracelet/charm/blob/main/.github/workflows/build.yml#L7)

bashbunni (approved):
Nice! Thanks for tagging me to review, it's always helpful :)

examples/simple/main_test.go
bashbunni, line 1: There's another PR (#759 ) where this test isn't passing, but it passes here just fine

.github/workflows/build.yml
caarlos0: not sure we can: the meta workflow just clone, build and test `./...`, we also do the testing in examples and tutorials... maybe if we change it to allow to pass the working directory 🤔 

anyway, will have a look on it in another PR 👍🏻

examples/simple/main_test.go
caarlos0: 🤔

Issue #573, opened 2022-10-24
Altscreen: rendering artifacts when resizing window


purpleclay:
Raising this issue based on the discussion with @muesli #544

It appears that with some terminals if you use `lipgloss` to fill available space with a background colour it automatically pads with whitespace. This causes a rendering glitch when you resize the terminal window due to word wrapping. A terminal such as Alacritty doesn't exhibit this behaviour. From testing, it exhibits this behaviour with iTerm2 and the native Mac Terminal.app.

https://user-images.githubusercontent.com/106762954/196023116-87c03990-5678-4c9e-95f7-46c6dbfcb074.mov

A potential solution is to fill the available height of the terminal with empty lines to prevent this wrapping behaviour.

Example App:

```go
package main

import (
	"log"
	"strings"

	tea "github.com/charmbracelet/bubbletea"
	"github.com/charmbracelet/lipgloss"
)

var (
	PrimaryColour = lipgloss.Color("#3a1577")

	Banner = lipgloss.NewStyle().Background(PrimaryColour).Height(2)
)

type model struct {
	width int
	text  string
}

func (m model) Init() tea.Cmd {
	return nil
}

func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	cmds := make([]tea.Cmd, 0)

	switch msg := msg.(type) {
	case tea.WindowSizeMsg:
		m.width = msg.Width
	case tea.KeyMsg:
		switch msg.String() {
		case "ctrl+c":
			return m, tea.Quit
		}
	}

	return m, tea.Batch(cmds...)
}

func (m model) View() string {
	var b strings.Builder

	b.WriteString(Banner.Copy().Width(m.width).Render(m.text))
	return b.String()
}

func main() {
	m := model{
		text: "hello!",
	}

	if err := tea.NewProgram(m, tea.WithAltScreen()).Start(); err != nil {
		log.Fatal(err)
	}
}
```

kristiandrucker:
I'm having the same issue and am just using viewort without lipgloss. Could this be an underlying issue in the renderer?

meowgorithm:
This is indeed an improvement that can be made to the renderer: thanks for flagging the issue. In the meantime, you can workaround the issue by manually clearing the screen on resize.

```diff
diff --git a/main.go b/main.go
index 426c6cb..a777967 100644
--- a/main.go
+++ b/main.go
[...]

Issue #717, opened 2023-04-09
Inline: rendering artifacts when resizing window


StarpTech:
**Describe the bug**

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS [e.g. Ubuntu, Mac OS]: PopOS (Ubuntu)
 - Shell [e.g. zsh, fish]: zsh
 - Terminal Emulator [e.g. kitty, iterm]: no
 - Terminal Multiplexer [e.g. tmux]: no

**To Reproduce**
1. Run fancy-list example
2. Resize the terminal multiple times to a very minimal height

**Source Code**
[Please include source code if needed to reproduce the behaviour. 
](https://github.com/charmbracelet/bubbletea/tree/master/examples/list-fancy)

**Expected behavior**
The program should not repaint the program on multiple lines.

**Screenshots**
[Screencast_from_04-09-2023_034642_PM.webm](https://user-images.githubusercontent.com/1764424/230777720-41efe195-75e3-4412-998e-0dcc36f69fb5.webm)

**Additional context**
Replicated in Gnome Terminal, JetBrains In-built, VsCode

meowgorithm:
Thanks for the report! This is a dup of #573—closing this one accordingly.

meowgorithm:
Reopening per the discussion in #573 so we can track inline rendering separately from the altscreen efforts.

StarpTech:
@meowgorithm friendly ping.

meowgorithm:
> I'm curious, what's the difference between bubbletea inline render and clearing, drawing the text manually? I can't reproduce any visual bugs when printing and updating libpgloss text to stdout directly.

Lip Gloss merely prints to the terminal and exits, whereas Bubble Tea manages the cursor and optimizes redraws to some degree. So when the terminal resizes, Bubble Tea needs to redraw and update it’s renderer’s state, whereas text printed with Lip Gloss remains solely in the jurisdiction of the terminal’s own renderer.

PR #738, opened 2023-05-11
fix: always repaint during a resize

meowgorithm:
Fixes https://github.com/charmbracelet/bubbletea/issues/573, https://github.com/charmbracelet/bubbletea/issues/717
Answers https://github.com/charmbracelet/bubbletea/discussions/544


meowgorithm:
Alas, this will actually need work closer to the renderer itself.

PR #740, opened 2023-05-18
fix: Possible Goroutine hang if input occurs during program exit

gabe565:
This only seems to happen in some cases, but I have noticed if I use bubbletea with the `tea.WithInput` option there's a chance `readLoop` will hang on line 92 if a user inputs characters at the same time as the program exits. This results in a very gradual memory leak.

This PR allows `readLoop` to short-circuit when reading input if the program has been killed.

### DIFF
--- a/tty.go
+++ b/tty.go
@@ -89,7 +89,11 @@ func (p *Program) readLoop() {
 		}
 
 		for _, msg := range msgs {
-			p.msgs <- msg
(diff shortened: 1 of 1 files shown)


gabe565:
This was resolved in b1e7f42ab0625c298c70f3dcd1f04da979a2e8be

Issue #741, opened 2023-05-20
bubbletea app hangs when quitting about 10% of the time.


collinvandyck:
**Describe the bug**
When my model returns `tea.Quit`, sometimes, maybe 10% of the time, it won't actually quit. I was able to dump the goroutine stacks when it was hanging by sending it the ABRT signal and saw that it was blocking on this line:

https://github.com/charmbracelet/bubbletea/blob/d1a16bd88370daba01ea34b9007e050f56adc97c/standard_renderer.go#L99

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS [MacOS Ventura]
 - Shell [zsh]
 - Terminal Emulator [kitty]
 - Terminal Multiplexer [tmux]

**To Reproduce**

Run the following program. As soon as it starts, you can hit q to quit it. You might need to do this 10-20 times to get it to eventually hang when quitting.

> go.mod

```go
module gol

go 1.20

require (
	github.com/aymanbagabas/go-osc52/v2 v2.0.1 // indirect
	github.com/charmbracelet/bubbletea v0.24.0 // indirect
	github.com/containerd/console v1.0.4-0.20230313162750-1ae8d489ac81 // indirect
	github.com/lucasb-eyer/go-colorful v1.2.0 // indirect
	github.com/mattn/go-isatty v0.0.18 // indirect
	github.com/mattn/go-localereader v0.0.1 // indirect
	github.com/mattn/go-runewidth v0.0.14 // indirect
	github.com/muesli/ansi v0.0.0-20211018074035-2e021307bc4b // indirect
	github.com/muesli/cancelreader v0.2.2 // indirect
	github.com/muesli/reflow v0.3.0 // indirect
	github.com/muesli/termenv v0.15.1 // indirect
	github.com/rivo/uniseg v0.2.0 // indirect
	golang.org/x/sync v0.1.0 // indirect
	golang.org/x/sys v0.6.0 // indirect
	golang.org/x/term v0.6.0 // indirect
	golang.org/x/text v0.3.8 // indirect
)
```

> main.go
```go
package main

import (
	"bytes"
	"math/rand"
	"time"

	tea "github.com/charmbracelet/bubbletea"
)

const fps = 20
const alive = "😂"
const dead = "🌊"

func main() {
	model := model{}
	p := tea.NewProgram(model)
	p.Run()
}

type board [][]bool

type model struct {
	board      board
	ready      bool
	lastUpdate time.Time
	ticked     bool
[...]

PR #742, opened 2023-05-20
Fix race condition in standard renderer.

collinvandyck:
Issue: https://github.com/charmbracelet/bubbletea/issues/741

When stopping the program, it would occasionally hang. Looking at the stacktrace, it is because the listen goroutine is trying to flush, which needs the mutext, but because stop() also held the mutex, it would deadlock.

To address this we make lock aquisition a little more granular and break out the line clearing into a separate method which grabs and then releases the mutext.


muesli:
I think we might just want to move closing the channel above the final flush and clear line. Preparing a PR to test.

collinvandyck:
Ah I like the you PR much better. thanks for taking a look!

collinvandyck:
Fixed in https://github.com/charmbracelet/bubbletea/pull/757

PR #747, opened 2023-05-26
.github: switch to using renovate batched update prs

paralin:
This is a suggestion to improve dependency upgrade PRs for bubbletea.

The provided renovate.json configuration batches updates into a single PR that the bot keeps up to date with the latest versions of deps.

This helps to prevent pull request spam from dependabot + PR fatigue.

Here's an example of what an upgrade PR looks like: https://github.com/aperturerobotics/controllerbus/pull/104

Also keeps the github workflow steps up to date.


caarlos0:
does it work for action updates, too?

if so, I love this!

caarlos0:
answering my own question: yes it does

meowgorithm:
@caarlos0 and @muesli: does this one still make sense?

caarlos0:
i think we can close this tbh, we're already using dependabot everywhere...

meowgorithm:
Just wanted to say appreciate the PR (as always) @paralin. In this case we're sticking with our existing tooling because of the degree in which it's woven into our workflows.

paralin:
That's fine, it was just a suggestion. That said, I still think the advantages to batched PRs are there.

meowgorithm:
It's a fair point; will definitely keep it mind when we take a breath and reexamine our tooling.

PR #763, opened 2023-06-14
Add typioca to "Bubble Tea in the Wild" section

bloznelis:
:bubble_tea:


meowgorithm:
Welp, that's a pretty apt name for a Bubble Tea app. Thanks for the PR!

PR #770, opened 2023-06-29
fix: Index out of range in examples/credit-card-form when ccn is empty

grafviktor:
Fixes Index out of range error in examples/credit-card-form when credit card number(CCN) is not provided.

To reproduce:

- Run examples/credit-card-form/main.go
- Do not type anything in "Card number"
- Press UP or DOWN keys on your keboard (or press TAB several times)

_ccnValidator_ will trigger and will attempt to read a symbol which is located at _len("")-1_, that will cause Index out of range error (because CCN is empty).


maaslalani:
Thanks so much!

PR #772, opened 2023-07-06
chore(examples): improved model initialization consistancy

samyosm:
I have made sure to maintain consistency across all examples by aligning them with the documentation, which uses an `initialModel` method for model initialization.


muesli:
While I prefer the naming consistency (`newModel` -> `initialModel`), we should probably simplify the code as much as possible and get rid of the method in some examples. We're also planning to streamline the bubbles a bit and make them proper `tea.Models`.

PR #773, opened 2023-07-06
build: group dependabot updates

caarlos0:


muesli (approved):
LGTM, but I have no experience with this new(?) syntax. Let's try it out? :-)

PR #778, opened 2023-07-07
chore: various linter-related improvements and overrides

meowgorithm:
Various linter-suggested improvements, and some overrides.

Also, apologies @caarlos0: I completely overlooked #724 back in April. Because input parsing has changed so much since then I figure it's easier at this point to simply supersede your earlier PR (but let me know if you feel differently).

Supersedes #724.

### DIFF
--- a/logging.go
+++ b/logging.go
@@ -1,6 +1,7 @@
 package tea
 
 import (
+	"fmt"
 	"io"
 	"log"
 	"os"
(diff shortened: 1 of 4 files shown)


maaslalani (approved):

caarlos0 (approved):

caarlos0:
> Also, apologies @caarlos0: I completely overlooked #724 back in April. Because input parsing has changed so much since then I figure it's easier at this point so simply supersede your earlier PR (but let me know if you feel differently).

no problem, your changes are actually better than mine (and I also had forgotten about that PR hehe)

logging.go
```diff
@@ -32,9 +33,9 @@ type LogOptionsSetter interface {
 
 // LogToFileWith does allows to call LogToFile with a custom LogOptionsSetter.
 func LogToFileWith(path string, prefix string, log LogOptionsSetter) (*os.File, error) {
-	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0o644)
+	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0o644) //nolint:gomnd
```
muesli, line 36: Actually, this might be a valid warning and `0o600` could be more sensible for (potentially) sensitive data like a logfile?

meowgorithm: Good call, I think that makes sense for this use case.

Regardless, this is still a magic number warning. I had originally pulled the number into a const but found trying to name a token describing `0o644` less clear than the literal representation. So, regardless, I’d probably keep the `nomnd` directive.

PR #791, opened 2023-08-01
fix: race changing ignoreSignals

caarlos0:
multiple goroutines can try  to `ReleaseTerminal` (and/or `RestoreTerminal`) at the same time, causing a data race.

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -154,7 +154,7 @@ type Program struct {
 
 	// was the altscreen active before releasing the terminal?
 	altScreenWasActive bool
-	ignoreSignals      atomic.Bool
+	ignoreSignals      uint32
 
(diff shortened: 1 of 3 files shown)


aymanbagabas (approved):

tea.go
```diff
@@ -153,7 +154,7 @@ type Program struct {
 
 	// was the altscreen active before releasing the terminal?
 	altScreenWasActive bool
-	ignoreSignals      bool
+	ignoreSignals      atomic.Bool
```
aymanbagabas, line 157: `atomic.Bool` was only introduced in go1.19, use `uint32` and `atomic.StoreUint32` instead

caarlos0: Indeed!

aymanbagabas (approved):

PR #796, opened 2023-08-09
feat: WithoutEmptyRenders

caarlos0:
experimental


meowgorithm:
I _think_ no need at the moment. Worst case we have the history here and we can re-open (or re-implement) if we need it.

PR #811, opened 2023-08-23
fix(render): set and unset RUNEWIDTH_EASTASIAN in program lifetime

bashbunni:
I still need to test this before merging, just need to find a good candidate for when `RUNEWIDTH_EASTASIAN` fixes a rendering issue. Also should decide what we want to do when we error out since our logger seems to introduce new deps we might not want?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/charmbracelet/bubbletea
 go 1.18
 
 require (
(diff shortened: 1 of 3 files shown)


meowgorithm:
Closing this per our discussion; we're basically moving over to `rivo/uniseg` across the board.

PR #812, opened 2023-08-24
refactor: remove redundant nil check in `flush`

Juneezee:
From the Go docs:

> "If the map is nil, the number of iterations is 0." https://go.dev/ref/spec#For_range

Therefore, an additional nil check for before the loop is unnecessary. Example: https://go.dev/play/p/zQyUnddPFbl


muesli:
Well spotted!

caarlos0 (approved):

PR #838, opened 2023-10-10
Lip Gloss table resizing demo

maaslalani:
Example adds a Lip Gloss table rendered with height and width set to the screen.

### DIFF
--- a/examples/table-resize/main.go
+++ b/examples/table-resize/main.go
@@ -1,0 +1,136 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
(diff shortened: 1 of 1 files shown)


meowgorithm:
This lgtm—but it looks like we need to update the build matrix for Ubuntu since `lipgloss` is on `1.18` now.

meowgorithm:
Looks like this doesn't compile anymore (I think because the table API has since changed). @maaslalani do you mind updating accordingly so we can get this in?

maaslalani:
Fixed!

PR #839, opened 2023-10-11
handle "q" for quitting in list-simple example

sharunkumar:


maaslalani:
Thanks so much for the fix! We should definitely handle `q` since it's in the help of the example! 😄

Issue #680, opened 2023-03-02
Incorrect order when nesting `Sequence` commands


smac89:
When a sequence command is nested, `bubbletea` does not wait for the nested sequence to finish before it executes the sibling command.

Bug can be fully reproduced by cloning and running the code at https://github.com/MVCE-Superstars/bubbletea

Changing line 58 in main.go, to nest a `Batch` or simply removing the nested sequence and replace it with just the function it contains, seems to produce the desired effect.

PR #843, opened 2023-10-15
fix: process nested sequence messages in order

Hrily:
Fixes: #680

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -384,6 +384,14 @@ func (p *Program) eventLoop(model Model, cmds chan Cmd) (Model, error) {
 							continue
 						}
 
+						// process nested sequence messages in order
+						if sequenceMsg, ok := msg.(sequenceMsg); ok {
+							for _, cmd := range sequenceMsg {
+								p.Send(cmd())
+							}
+							continue
+						}
+
 						p.Send(msg)
 					}
 				}()


Hrily:
[hrily/bubbletea/examples/nested-sequence](https://github.com/Hrily/bubbletea/blob/hrily/nested-seq-example/examples/nested-sequence/main.go) is the example to reproduce the bug from the issue and gives the expected output with the fix from this commit:

```


    Searching for something...
  ──────────────────────────────




  Something is not loaded!
  Something is loaded!
```

tea.go
```diff
@@ -384,6 +384,14 @@ func (p *Program) eventLoop(model Model, cmds chan Cmd) (Model, error) {
 							continue
 						}
 
+						// process nested sequence messages in order
+						if sequenceMsg, ok := msg.(sequenceMsg); ok {
+							for _, cmd := range sequenceMsg {
+								p.Send(cmd())
```
CalebJohnHunt, line 390: What if we have deeply nested sequences (3+)? Imagine we have a `cmd` like:
```
sequenceMsg: [
    sequenceMsg: [
        sequenceMsg: [customMsg11, customMsg12],
        sequenceMsg: [customMsg21, customMsg22]
    ]
]
```
where each `Msg` represents a `cmd` and what will be returned after running it.

Doesn't this solution only work for the first two sequences? The outermost sequence is handled by line 363, and then the second sequence is handled here, but the third and fourth sequences (with the `customMsg*`s) will be done concurrently -- `customMsg11` will happen before `customMsg12`, but `customMsg21` could occur before `customMsg11`, after `customMsg11`, or after `customMsg12`. Am I missing something?

Issue #234, opened 2022-02-18
Recover panics from within cmds goroutines


brooks-connor-a:
Panics are handled by recovering the panic and gracefully tearing down the Program. However goroutines started within the Program are outside the scope of the recover, and as such can end in the program terminating without properly tearing down. This can result in the terminal getting stuck in a non-interactive state until it is reset.

Here's a minimal reproduction:
```
package main
​
import ( 
    tea "github.com/charmbracelet/bubbletea"
)
​
type model struct {}

func (m model) Init() tea.Cmd {
    return nil
}

func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) { 
    return m, func() tea.Msg { 
        if true { 
            panic("")
        } 
        return tea.KeyEnter
    } 
}
​
func (m model) View() string { 
    return ""
}
​
func main() { 
    p := tea.NewProgram(model{})
    p.Start()
}
```

decentral1se:
Thanks for reporting @brooks-connor-a, ran into this also.

joerdav:
I took a look at this, it seems like the best way to solve was to recover the panic and write to the `errs` channel. Happy for some feedback if it's not in the spirit of how bubbletea works.

meowgorithm:
Hi! Just a note that this is now fixed and available in [Bubble Tea v1.1.1](https://github.com/charmbracelet/bubbletea/releases/tag/v1.1.1). You can use `tea.Kill` to reset the terminal from panics externally.

```go
p := tea.NewProgram(model{})

go func() {
	time.Sleep(3 * time.Second)
	defer p.Kill()
	panic("Oh no!")
}()
	
if _, err := p.Run() {
	log.Fatal(err)
}
```

PR #846, opened 2023-10-18
fix: gracefully quit panicking Cmd

joerdav:
Fixes: #234

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -292,14 +292,7 @@ func (p *Program) handleCommands(cmds chan Cmd) chan struct{} {
 				go func() {
 					// Recover from panics.
 					if !p.startupOptions.has(withoutCatchPanics) {
-						defer func() {
-							if r := recover(); r != nil {
-								fmt.Printf("Caught panic:\n\n%s\n\nRestoring terminal...\n\n", r)
-								debug.PrintStack()
-								p.errs <- fmt.Errorf("%v", r)
-								return
-							}
-						}()
+						defer p.recoverFromPanic()
 					}
 					msg := cmd() // this can be long.
 					p.Send(msg)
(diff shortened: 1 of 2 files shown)


muesli:
We should try and add a test for this.

blvrd:
I'd be happy to help get this over the finish line if there's anything else that needs to be done. It would be nice to not need to `reset` the terminal if there's a panic inside a `tea.Cmd`.

meowgorithm:
@blvrd can you verify that it fixes #1127?

blvrd:
No, it doesn't fix #1127 since the panic in the reproduction script is happening outside of Bubbletea. If the `recoverFromPanic` method from this PR was exported, you _could_ fix #1127 with this approach:

```go
package main

import (
	"fmt"
	"os"
	"time"

)

type model struct{}

	return nil
}

	return m, nil
}

	return "I will panic"
}

func main() {
	m := model{}
        p := tea.NewProgram(m)

	go func() {
                defer p.RecoverFromPanic() // recover here
		time.Sleep(2 * time.Second)
		panic("")
	}()

		fmt.Println("Uh oh, we encountered an error:", err)
		os.Exit(1)
	}
}
```

This doesn't feel like a great solution though. Maybe you'd just have to be responsible for your own panic recovery for goroutines that execute outside of the scope of the `tea.Program`?

tea.go
```diff
@@ -300,9 +304,22 @@ func (p *Program) handleCommands(cmds chan Cmd) chan struct{} {
 	return ch
 }
 
+func (p *Program) recoverFromPanic() {
+	if r := recover(); r != nil {
+		fmt.Printf("Caught panic:\n\n%s\n\nRestoring terminal...\n\n", r)
+		debug.PrintStack()
+		p.errs <- fmt.Errorf("%v", r)
+		return
+	}
```
meowgorithm, line 313: I believe this is very close to code used in `Program.Run` so we should probably use this method there as well. That said, this method appears to be missing the `p.shutdown(true)` we see in `Program.Run`: is that intentional?

blvrd: cc: @joerdav

meowgorithm:
Just a heads up that we're fixing this over at #1134; it's nuanced and we want to be very deliberate with this one. @joerdav, we appreciate you flagging this one!

meowgorithm:
Hello again! Just a note that the fix for this is now available in Bubble Tea v1.1.1.

Issue #847, opened 2023-10-23
Execute multiply nested SequenceMsg and BatchMsg in the proper order.


wolfmagnate:
# **Is your feature request related to a problem? Please describe.**

## Description
As stated in issue #680, the current `eventLoop` method handling for `sequenceMsg` executes only once nested `BatchMsg` in the proper order, but nested `sequenceMsg` and multiple nested `BatchMsg` or `sequenceMsg` will not follow the specified order.
This is due to the fact that the `eventLoop` method calls `p.Send` without checking the result of the message that is the result of the call to the command contained in `sequenceMsg` all the way through the nest.

## Example
The following shows the desired and actual calling order for a given `sequenceMsg`.

```go
func (m model) Init() tea.Cmd {
	return tea.Sequence(
		// Command1
		tea.Batch(
			// Command1-1
			tea.Sequence(
				tea.Println("1-1-1"),
				tea.Println("1-1-2")
			),
			// Command1-2
			tea.Batch(
				tea.Println("1-2-1"),
				tea.Println("1-2-2")
			)
		),
		// Command2
		tea.Println("2"),
		// Command3
		tea.Sequence(
			// Command3-1
			tea.Batch(
				tea.Println("3-1-1"),
				tea.Println("3-1-2")
			),
			// Command3-2
			tea.Sequence(
				tea.Println("3-2-1"),
				tea.Println("3-2-2")
			)
		),
		// last Command
		tea.Quit,
	)
}
```

These are important points of the example above.
- Command1 is generated by `tea.Batch`, so Command1-1 and Command1-2 are executed in parallel.
- Command1-1 is generated by `tea.Sequence`, so Command1-1-1 and Command1-1-2 are executed in sequence.
- Command1-2 is generated by `tea.Batch`, so Command1-2-1 and Command1-2-2 are executed in parallel.
- Command2 is executed after all commands executed by Command1 (1-1-1, 1-1-2, 1-2-1, 1-2-2) have been executed.
- Command3 is executed after Command2 finishes.
- Command3 is generated by `tea.Sequence`, so Command3-1 and Command3-2 are executed in sequence.
- Command3-1 is generated by `tea.Batch`, so Command3-1-1 and Command3-1-2 are executed simultaneously.
- Command3-2 is generated by `tea.Sequence`, so Command3-2-1 and Command3-2-2 are executed in sequence.
- Quit is executed after all these commands are completed.

This is a simple sequence diagram for Command1.
[...]

PR #848, opened 2023-10-23
feat: properly call nested sequenceMsg and batchMsg

wolfmagnate:
# what this PR does
fix problems described in #847 

# problems
I have to check...
- method name: execSequenceMsg and execBatchMsg are good names for these methods?
- bugs: Does this change make other problems?
- optimization: Is there any performance problems when I handle nested messages with recursive functions?

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -355,13 +355,12 @@ func (p *Program) eventLoop(model Model, cmds chan Cmd) (Model, error) {
 				p.exec(msg.cmd, msg.fn)
 
 			case BatchMsg:
-				for _, cmd := range msg {
-					cmds <- cmd
-				}
+				go p.execBatchMsg(msg)
 				continue
 
 			case sequenceMsg:
 				go p.execSequenceMsg(msg)
+				continue
 			}
 
 			// Process internal messages for the renderer.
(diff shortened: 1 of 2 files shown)


3v1n0:
Any hope in getting this handled? :)

tea.go
```diff
@@ -402,6 +376,50 @@ func (p *Program) eventLoop(model Model, cmds chan Cmd) (Model, error) {
...
+				p.execSequenceMsg(msg)
+			default:
+				p.Send(msg)
+			}
+			return nil
+		})
+	}
+
+	//nolint:errcheck
+	g.Wait() // wait for all commands from batch msg to finish
```
aymanbagabas, line 420: Since we're not using or returning errors, I would use a waitgroup instead

wolfmagnate:
Thank you for your review. Since this PR is too old and has a conflict, I would write the same code-fix for the current code.

tea.go
```diff
@@ -471,60 +471,12 @@ func (p *Program) eventLoop(model Model, cmds chan Cmd) (Model, error) {
...
-							p.Send(msg)
-						}
-					}
-				}()
-
-			case setWindowTitleMsg:
-				p.SetWindowTitle(string(msg))
-
-			case windowSizeMsg:
-				go p.checkResize()
```
merhard, line 527: Why were these handled messages removed? This seems to have led to a regression in `v1.3.9` for the `tea.SetWindowTitle(title)` and `tea.WindowSize()` commands.

PR #853, opened 2023-11-05
fix(examples): run filepicker on stdout (versus stderr)

### LINKED ISSUE #852: Wrong Output prevents file selection from working properly when multiple panes are used
https://github.com/charmbracelet/bubbletea/blob/5536bca34e218a1e17a55e5ba36c22fab5784f44/examples/file-picker/main.go#L91C40-L91C40

Please see discussion here: https://discord.com/channels/1032368467246075914/1170385465204035774/1170385465204035774

taigrr:
Fixes #852 

The cursor has trouble mapping to the correct location when using stderr. Using stdout is more reliable, and this example isn't taking advantage of stderr in any meaningful way over stdout, so it's probably better to show people a more common example.

### DIFF
--- a/examples/file-picker/main.go
+++ b/examples/file-picker/main.go
@@ -88,7 +88,7 @@ func main() {
 	m := model{
 		filepicker: fp,
 	}
-	tm, _ := tea.NewProgram(&m, tea.WithOutput(os.Stderr)).Run()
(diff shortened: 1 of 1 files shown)


meowgorithm:
LGTM—and I agree.

PR #865, opened 2023-11-11
docs: add nom rss reader to bubble-tea-in-the-wild

guyfedwards:
Thanks for all the great libraries 🎉


meowgorithm:
Thanks for the PR — nice work!

PR #875, opened 2023-11-28
chore: optimize batches with one item

systay:
Minimal fix that returns the single command instead of wrapping it in a `Batch`


meowgorithm:
Awesome: this is very acute.

PR #885, opened 2023-12-14
chore(docs): add end_of_eden and CRT to Bubbletea in the Wild

BigJk:
Would be great a honour to have my 2 projects that are using bubbletea and other charm libs to the docs 😄


meowgorithm:
These are so great. Thank you for the PR!

PR #887, opened 2023-12-25
feat(wasm): Add empty functions for js builds to enable WASM builds

BigJk:
Hello charm `ヾ(＾∇＾)`

In this write-up I laid out which changes would be needed to make bubbletea easily compile for WASM: https://github.com/BigJk/bubbletea-in-wasm

One of the changes would be to introduce some simple *filler* functions for js inside bubbletea so that a build with `GOOS=js GOARCH=wasm` won't break. These are contained in this pull request 😀 Additionally 2 dependencies needs to be bumped but I didn't include it into this PR atm as this depends on the response in https://github.com/atotto/clipboard/pull/48

I still opened the PR to track the progress and will gladly answer any questions!


maaslalani:
Hi! Thanks so much for the PR! I believe this is a duplicate of https://github.com/charmbracelet/bubbletea/pull/748

cwarden:
Can we move this forward by using a fork of the clipboard package?

Dunkansdk:
Hello! Can we move this forward? Thanks!

PR #893, opened 2024-01-09
Fix grammar in README

kevgo:


meowgorithm:
Good catch, thank you!

PR #897, opened 2024-01-10
feat: reduce console/term dependencies

aymanbagabas:
Replace mattn/isatty and containerd/console with golang.org/x/term.

This mostly affects Windows. On Windows, unlike Unix, the console (TTY) has different handles for input/output. Using the Console API, we need to enable VT input on the input handle (CONIN) and VT processing on the output handle (CONOUT). Doing so enables processing VT sequences on Windows i.e. ANSI colors, mouse sequences, cursor movements, etc.

We already handle enabling VT processing for the program output using Termenv `EnableVirtualTerminalProcessing`. For the input side, we enable VT input right before setting the console to raw.

By doing this, we can drop both containerd/console and mattn/isatty.

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -21,11 +21,10 @@ import (
 	"sync/atomic"
 	"syscall"
 
-	"github.com/containerd/console"
-	isatty "github.com/mattn/go-isatty"
 	"github.com/muesli/cancelreader"
 	"github.com/muesli/termenv"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/term"
 )
 
 // ErrProgramKilled is returned by [Program.Run] when the program got killed.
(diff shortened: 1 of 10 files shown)


caarlos0 (approved):

meowgorithm:
Welp, this is awesome. Have we tested on Windows?

muesli:
We should also very carefully test this in combination with `ExecProcess`, which needs to reset & restore terminal state.

tea.go
```diff
@@ -147,24 +146,17 @@ type Program struct {
 	renderer      renderer
 
 	// where to read inputs from, this will usually be os.Stdin.
-	input        io.Reader
+	input io.Reader
+	// tty is null if input is not a TTY.
+	tty          *os.File
+	ttyState     *term.State
```
caarlos0, line 152: this could prob be called `previousTtyState` or something like that, since it is the state the term had before `MakeRaw` 🤔

meowgorithm (approved):

PR #905, opened 2024-01-17
Prevent the renderer from hanging if the main model panics during its init

### LINKED ISSUE #904: Application hang when calling panic
**Describe the bug**
I have gotten my application to hang in a way that I can't get out of it without killing it from another terminal when calling panic while the application is initializing. If the application panics at a later time it exits properly.

**Setup**
 - OS - macOS Sonoma 14.2.1
 - Shell - zsh
 - Terminal Emulator - iterm2
 - Terminal Multiplexer - N/A

**To Reproduce**
1. Clone https://github.com/timmattison/bubbletea-hang
2. Run `go run main2.go model.go`
3. Kill the process from another terminal

**Source Code**
Source code is in linked repo

**Expected behavior**
I would expect that it panics and exits but instead gets stuck when trying to write to the done channel.
 https://github.com/charmbracelet/bubbletea/blob/8f3464a75600e991bbce22229d6e5b99975416f0/standard_renderer.go#L121

timmattison:
This should fix #904

The issue arises if the main model panics during init. If this happens it tries to send a signal via `r.done`. But nothing is reading `r.done` until `go r.listen()` is called in `r.start()`.

In `r.start()` the `r.ticker` value is set up, then `r.once` is initialized, and then finally `go r.listen()` is called. Since `r.ticker` is only assigned a value in this one place we can be reasonably certain that if it is non-nil that `go r.listen()` has been called and it is monitoring the `r.done` channel.

An alternate solution would be to add a new variable to indicate that the system has entered `r.listen()`. I like that solution better, but I wasn't sure if it makes sense to add another variable for this edge case.

### DIFF
--- a/standard_renderer.go
+++ b/standard_renderer.go
@@ -95,9 +95,7 @@ func (r *standardRenderer) start() {
 // stop permanently halts the renderer, rendering the final frame.
 func (r *standardRenderer) stop() {
 	// Stop the renderer before acquiring the mutex to avoid a deadlock.
-	r.once.Do(func() {
-		r.done <- struct{}{}
-	})
+	r.sendDoneSignal()
 
 	// flush locks the mutex
 	r.flush()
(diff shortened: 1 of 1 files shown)


meowgorithm:
Hi! Does #926 fix this? If so, would prefer that one as it’s a simpler change.

timmattison:
It looks like it does but I haven't tested it. I prefer that one too if it works.

meowgorithm:
Just checked and it appears to. Speaking of which, thank you for the very well documented issue; made it very easy to vet.

maaslalani:
Closing this PR as it seems that https://github.com/charmbracelet/bubbletea/pull/926 solves the issue.

Please feel free to re-open this PR if #926 doesn't solve #904 / #924 after it's merged.

PR #908, opened 2024-01-21
chore(docs): remove the tea.EnterAltScreen from Init

timmattison:
The docs say you shouldn't do this - https://github.com/charmbracelet/bubbletea/blob/8f3464a75600e991bbce22229d6e5b99975416f0/screen.go#L29 - and it works just fine with it removed.

Removing it will make sure that people who are new don't go down the wrong path looking at this example.


meowgorithm:
Great catch. Thanks for another acute PR!

PR #911, opened 2024-01-24
fix: only print control chars if output is a tty

caarlos0:
Usually doing it anyway is not a problem, but if you run `:!your-program` on Vim/neovim, it'll print a lot of control chars.

This kinda hammers it to not do so if the output is not a TTY, but I'm not sure what the side effects of this are... my understanding is that this would be a problem only if the terminal reports it is not a tty, but properly handles the CSI seqs anyway... 🤔

refs https://github.com/charmbracelet/mods/issues/98
refs https://github.com/charmbracelet/mods/issues/198

---

PS: expecting tests to fail, wanna make sure this is the right approach first

### DIFF
--- a/standard_renderer.go
+++ b/standard_renderer.go
@@ -4,10 +4,12 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


caarlos0:
Came up with a different fix in mods instead.

PR #926, opened 2024-02-14
fix: deadlock condition on model init panic

### LINKED ISSUE #924: Application deadlock when initial model's init panics
**Describe the bug**
With the current flow in `*Program.Run()`, the renderer isn't started until after the initial model has its `.Init()` function called. This currently causes an issue in the event that the initial model panics unexpectedly because the program is set to recover from panics, but can't finish shutting down because it's in a deadlock. I propose moving the `p.renderer.start()` to before the `p.initialModel.Init()` call in order to prevent this deadlock scenario. This will at least have the renderer start listening for a done signal on the context in the event of a panic.

I'm unsure if there are other unintended side effects from this change, but I don't see any from testing all of the examples.

**To Reproduce**
Running the code in the `Source Code` section will reproduce the issue. Instead of the expected panic being bubbled up, the app instead gets permanently stuck hanging waiting for
```go
	// Stop the renderer before acquiring the mutex to avoid a deadlock.
	r.once.Do(func() {
		r.done <- struct{}{}
	})
```
to execute.

**Source Code**
```go
package main

import tea "github.com/charmbracelet/bubbletea"

type model struct {
}

	//TODO implement me
	panic("implement me")
}

	//TODO implement me
	panic("implement me")
}

	//TODO implement me
	panic("implement me")
}

func main() {
	m := model{}
	if _, err := tea.NewProgram(m).Run(); err != nil {
		panic(err)
	}
}
```

**Expected behavior**
```bash
Caught panic:                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                    
implement me                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                    
Restoring terminal...   
```

**Additional context**
I have a fork with a branch created with the proposed change [here](https://github.com/eolso/bubbletea/tree/fix-init-deadlock).

eolso:
Fixes #924

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -515,6 +515,9 @@ func (p *Program) Run() (Model, error) {
 		p.renderer.enableMouseSGRMode()
 	}
 
+	// Start the renderer.
+	p.renderer.start()
+
 	// Initialize the program.
 	model := p.initialModel
 	if initCmd := model.Init(); initCmd != nil {
@@ -530,9 +533,6 @@ func (p *Program) Run() (Model, error) {
 			}
 		}()
 	}
-
-	// Start the renderer.
-	p.renderer.start()
 
 	// Render the initial view.
 	p.renderer.write(model.View())


meowgorithm:
Nice, fix seems good to me. @muesli whatcha think?

meowgorithm:
Supersedes https://github.com/charmbracelet/bubbletea/pull/905.
Fixes https://github.com/charmbracelet/bubbletea/pull/904.

muesli:
Still have to figure out how well this integrates with our existing render context proposals. In the current `master` it seems to be a reasonable change, so I'm gonna merge this one.

meowgorithm:
@muesli Gonna merge this one unless you feel differently. Fixes the issue in my testing, it logically makes sense to me, and it's a small adjustment to the source code.

muesli:
Yes, still looks good to me. I'll just click the button before I forget again :smile:

PR #936, opened 2024-03-02
feat: add support for Cmd wrapping (for routing)

JaredReisinger:
This is one possible solution for allowing components to wrap any `tea.Cmd` that comes from a sub-component's `Update()`.  See the comment in the code for other possible options.

Fixes #751

----

I'm creating this as a draft PR because this seemed like the easiest way to share one possible solution to the "how can a component reliably route a message to the correct subcomponent?" question.  There are pros/cons to any of the solutions; this just provides a starting point for the discussion.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -43,6 +43,66 @@ func Sequence(cmds ...Cmd) Cmd {
 // sequenceMsg is used internally to run the given commands in order.
 type sequenceMsg []Cmd
 
+// Wrap allows a parent component to accurately wrap the results of any Cmd,
+// regardless of whether it's a direct Cmd or a collective one like Batch or
+// Sequence.  Note that the "wrapping" arg could also be an interface{}, or a
+// generic type, or even a full `func(Cmd) Msg` (for the most flexibility), but
+// using an int seems to satisfy the 80:20 rule of the simplest solution for the
+// most-common case: enabling routing for sub-components.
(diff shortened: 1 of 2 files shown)


DavidS-ovm:
Would this also be helpful to reduce the custom ID handling in https://github.com/charmbracelet/bubbles/blob/master/spinner/spinner.go ?

JaredReisinger:
_Well..._ not necessarily.  Here's the problem: the way the spinner currently works, it requires _all_ parent components to pass _at least_ the `spinner.TickMsg` messages down to their child spinner(s).  The solution in the PR is aimed at the parent component/model, not the child.  So... it *could* help the parent to correctly delegate the `spinner.TickMsg` to the child spinner _without the parent having to know it was a `spinner.TickMsg`_.  This is a huge value to me as the author of a parent component, because it insulates me from internal changes of the child.  The child component adds an all-new message?  It's not even an exported type?  I don't care, I can blindly wrap the outcoming command and then get an incoming message that's flagged for a specific child.

Back to the spinner...  If (and only if) this is the _only_ way that a child spinner gets its message, then yes, the internal ID checking in spinner could probably go away.  But if I were the one maintaining spinner, I'd probably leave the ID handling in-place, because I wouldn't trust parent components to always do the "right thing" (e.g. use the solution in this PR), and I'd also want to keep some backward compatibility.

PR #958, opened 2024-03-16
Compact sequences like batches

jdhenke:
This change makes `Sequence` behave like `Batch` in how it handles `nil` entries, preventing a possible infinite loop.

## Background

I stumbled across this discrepancy when I noticed a bubbletea program I had written had CPU usage over 100% all the time, when I would expect it near zero as a CLI simply waiting for input.

I created a [simplified repro](https://github.com/jdhenke/repro-bubbletea-sequence-bug/blob/main/main.go), and noticed that swapping in `tea.Batch` for `tea.Sequence` fixed the CPU usage, which was nice, but it seems like either one _should_ work.

Digging in, it looks like `Batch` already had logic to handle all nils in #217 and more recently to return the only non-nil entry directly if possible in #875. This change reuses the same logic (and tests) for `Sequence` as well now, which avoids creating this infinite loop when using `Sequence` :+1: 

## Demo of Fix

Here's a walk through showing the fix in action within a go workspace.

```
$ cat go.work
go 1.21

toolchain go1.21.1

use (
	./bubbletea
	./repro
)
```

Here's showing the CPU issue before:

image

image

And here's showing it's fixed using this branch:

image

image

----

Let me know your thoughts and if you need this PR adjusted in any way, cheers.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -13,6 +13,27 @@ import (
 //		       return tea.Batch(someCommand, someOtherCommand)
 //	    }
 func Batch(cmds ...Cmd) Cmd {
+	return compactCmds[BatchMsg](cmds)
+}
+
+// BatchMsg is a message used to perform a bunch of commands concurrently with
+// no ordering guarantees. You can send a BatchMsg with Batch.
+type BatchMsg []Cmd
+
+// Sequence runs the given commands one at a time, in order. Contrast this with
+// Batch, which runs commands concurrently.
+func Sequence(cmds ...Cmd) Cmd {
+	return compactCmds[sequenceMsg](cmds)
+}
+
+// sequenceMsg is used internally to run the given commands in order.
+type sequenceMsg []Cmd
+
+// compactCmds ignores any nil commands in cmds, and returns the most direct
+// command possible. That is, considering the non-nil commands, if there are
+// none it returns nil, if there is exactly one it returns that command
+// directly, else it returns the non-nil commands as type T.
+func compactCmds[T ~[]Cmd](cmds []Cmd) Cmd {
 	var validCmds []Cmd //nolint:prealloc
 	for _, c := range cmds {
 		if c == nil {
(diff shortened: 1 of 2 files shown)


aymanbagabas (approved):

W-Floyd:
This might have caused a regression (or some other change to `Sequence`), in my case when returning a `Sequence` from a model `Init()`. Pinning to `v1.3.6` fixes it.

aymanbagabas:
> This might have caused a regression (or some other change to `Sequence`), in my case when returning a `Sequence` from a model `Init()`. Pinning to `v1.3.6` fixes it.

Could you please specify what's breaking when using `Sequence`?

W-Floyd:
I'm gonna investigate outside work hours and submit a proper bug report, as the only symptom I have so far is that the sequence simply doesn't get executed...

W-Floyd:
Update, bisected it to https://github.com/charmbracelet/bubbletea/pull/1463 instead, but can't yet figure out a minimal example.

W-Floyd:
I believe https://github.com/charmbracelet/bubbletea/pull/848 likely handles this. Commenting [this code](https://github.com/charmbracelet/bubbletea/blob/21eecd586367fd0cd78da6842c48f9c4b1185b6f/tea.go#L510-L517) at least fixes my use case (likely not the right long-term fix). For now, I can avoid nesting to work around this.

aymanbagabas:
> I believe #848 likely handles this. Commenting [this code](https://github.com/charmbracelet/bubbletea/blob/21eecd586367fd0cd78da6842c48f9c4b1185b6f/tea.go#L510-L517) at least fixes my use case (likely not the right long-term fix). For now, I can avoid nesting to work around this.

I've updated #848 with some changes and tests. Will merge and release once the CI pass.

W-Floyd:
Hmm... still the same issue - I'm trying to figure out a minimal example to test with.

W-Floyd:
<details>

<summary>Minimal example</summary>

Using `replace github.com/charmbracelet/bubbletea => ../bubbletea`

```go
package main

import (
	"fmt"
	"time"

)

var cnt = 10
var c = make(chan int, cnt)

type message int
type model int

func main() {
	tea.NewProgram(model(0)).Run()
}

func (m *model) seqMaker() tea.Cmd {
	return tea.Sequence(
		func() tea.Msg {
			return message(<-c)
		},
	)
}

	cmds := []tea.Cmd{}

	for i := 0; i < cnt; i++ {
		cmds = append(cmds, m.seqMaker())
		cmds = append(cmds, func() tea.Msg { return message(1) })
		go func() {
			time.Sleep(time.Millisecond * 500)
			c <- -1
		}()
	}

	return tea.Sequence(cmds...)
}

	switch msg := msg.(type) {
	case tea.KeyMsg:
		switch msg.String() {
		case "ctrl+c", "q":
			return m, tea.Quit
		}
	case message:
		m += model(msg)
		time.Sleep(time.Millisecond * 10)
		return m, nil
	}
	return m, nil
}

	return fmt.Sprintf("Count: %d\n", m)
}
```

</details>

# v1.3.6 (as expected)
out

# v1.3.7 - v1.3.9
out

aymanbagabas:
> Hmm... still the same issue - I'm trying to figure out a minimal example to test with.

Could it be a race condition? Like the program exits before the commands run?

W-Floyd:
(moved to https://github.com/charmbracelet/bubbletea/issues/1494)

PR #962, opened 2024-03-22
refactor: use x/term and x/ansi for renderer sequences

aymanbagabas:
This replaces termenv.Output with raw sequences and bytes.Buffer to handle vt screen and cursor events. Also uses `ansi.Truncate` instead of `reflow/truncate`


jmonroynieto:
The latest version of this package has graduated from experiments. The stable code should be `golang.org/x/term.` I haven't studied whether there might be breaking changes for the stable package. using an outdated experimental package related to #968

jmonroynieto:
It was my missunderstanding then. Closing.

caarlos0 (approved):

PR #965, opened 2024-03-25
refactor!: remove deprecated

aymanbagabas:
Remove deprecated commands in preparation for release


meowgorithm:
My only thought here is that we may want to hang onto the deprecated `Key` stuff for a little while longer in order to give users time to migrate as reworking `Key` is a substantial and fundamental change.

PR #983, opened 2024-04-19
docs: add goread rss/atom reader to the project list

TypicalAM:
Hi! Thanks for the amazing library!


aymanbagabas:
Hey @TypicalAM, please add goread to https://github.com/charm-and-friends/charm-in-the-wild instead as we moved the project list there 🙂

PR #986, opened 2024-04-21
chore: use errors.New to replace fmt.Errorf with no parameters will much better

ChengenH:
use errors.New to replace fmt.Errorf with no parameters will much better


Delta456 (approved):
`errors.New` is indeed better for this usage.

Issue #987, opened 2024-04-21
Allow WindowSizeMsg to be sent on sub-model Init


tylerolson:
When attempting to format content that is reliant on the width/height you must listen for `WindowSizeMsg` and store those values in the model:

```go
func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	var cmd tea.Cmd
	switch msg := msg.(type) {
	case tea.WindowSizeMsg:
		m.width = msg.Width
  		m.height = msg.Height
        }
}
```

This presents a challenge when switching to a different model, as the `WindowSizeMsg` msg is never sent again.

This means that you must either use global variables to keep track of the width/height, or pass around width/height variables to all of your sub models:

```go
func newOtherModel(width int, height int) otherModel {
	return otherModel{
		width:    width,
		height:   height,
	}
}
```

**Describe the solution you'd like**
Either allow a way to manually request the current WindowSize at say, Init(), or maybe provide a way to update the current model without using a mainmodel-submodels type architecture.

Maybe this is a non-issue for most, just something I found :)

meowgorithm:
Hi! So technically speaking you shouldn't use `Init` to initialize your model; its job is _just_ to return an initial command.

To pass the window width and height when switching to a new model you can simply set properties on the new model before returning it:

```go
func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	switch msg := msg.(type) {
	case tea.WindowSizeMsg:
		m.width = msg.Width
  		m.height = msg.Height
		return m, nil

        case switchModelsMsg:
		newModel := NewModel()
		newModel{width: m.width, height: m.height} // pass along width/height
		return newModel, newModel.Init() // switch models
        }
}
```

KevM:
In [bubbleo's navstack](https://github.com/KevM/bubbleo/blob/e8bf58588311a7d71c3c0335d46bd1abd867769b/navstack/model.go#L62) when pushing a new model on the stack I call Init and send the new model and Update with the window size msg for exactly the reason you mention. We need the model now in charge of the application to know what its size is.

meowgorithm:
@KevM Cool, so sounds like we're basically saying the same thing, right?

meowgorithm:
Reading this again, perhaps a `Cmd` for requesting the window size would indeed be useful.

tylerolson:
[...]

PR #988, opened 2024-04-22
feat: add a cmd to request window size

aymanbagabas:
Fixes: https://github.com/charmbracelet/bubbletea/issues/987#issuecomment-2068452658

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -191,3 +191,12 @@ func SetWindowTitle(title string) Cmd {
 		return setWindowTitleMsg(title)
 	}
 }
+
+type windowSizeMsg struct{}
+
+// WindowSize is a command that queries the terminal for its current size.
+func WindowSize() Cmd {
+	return func() Msg {
+		return windowSizeMsg{}
+	}
+}
--- a/commands.go
+++ b/commands.go
@@ -194,7 +194,11 @@ func SetWindowTitle(title string) Cmd {
 
 type windowSizeMsg struct{}
 
-// WindowSize is a command that queries the terminal for its current size.
+// WindowSize is a command that queries the terminal for its current size. It
+// delivers the results to Update via a [WindowSizeMsg]. Keep in mind that
+// WindowSizeMsgs will automatically be delivered to Update when the [Program]
+// starts and when the window dimensions change so in many cases you will not
+// need to explicitly invoke this command.
 func WindowSize() Cmd {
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

commands.go
```diff
@@ -191,3 +191,12 @@ func SetWindowTitle(title string) Cmd {
 		return setWindowTitleMsg(title)
 	}
 }
+
+type windowSizeMsg struct{}
+
+// WindowSize is a command that queries the terminal for its current size.
```
meowgorithm, line 197:
```suggestion
// WindowSize is a command that queries the terminal for its current size. It
// delivers the results to Update via a [WindowSizeMsg]. Keep in mind that
// WindowSizeMsgs will automatically be delivered to Update when the [Program]
// starts and when the window dimensions change so in many cases you will not
// need to explicitly invoke this command.
```

atticus-sullivan:
Because the PR got merged only a few hours ago and this issue is 100% certain linked to this PR, I'm writing no new issue but comment on this PR.

Sorry for waiting until after the merge before testing this. In my test, when I send a `tea.WindowSize()` in the `Update()` function of a model, it causes the program to hang up. My analysis shows that the reason behind this most probably (well tbh, I'm pretty certain on this) is because
you call `p.checkResize()` in the `eventLoop()` which in turn calls `p.Send()` which in the end sends on the `p.msgs` channel. The problem is (by default) this channel is not buffered, thus the send blocks until another goroutine reads from it (but the one which would read from it is the one sending on that channel).

I'm not too familiar with the codebase but searching for other occurrences of `p.Send()`, it seems like this function is called in a new goroutine usually. In this case I'm not sure if the `p.checkResize()` should call `go p.Send()` (would also affect the init stuff which I'm not familiar with) or if we should call `go p.checkResize()` in the `eventLoop()`.

I hope I didn't formulate this too complicated otherwise feel free to ask me what I actually mean. Any thoughts on this? Also, didn't someone else had this issue as well?

aymanbagabas:
@atticus-sullivan Thank you for noticing this one! I've opened https://github.com/charmbracelet/bubbletea/pull/1059 to fix this one and add an example to request window size.

atticus-sullivan:
Alright, thanks for the fast reply + fix. Also having an example for this is quite nice :+1:

PR #992, opened 2024-04-29
docs: add pug to "in the wild"

leg100:
Pug is a task manager for terraform: https://github.com/leg100/pug

It's been refreshing working with bubbletea, thank you for all your hard work.


meowgorithm:
This looks _so_ fun. Thanks for the PR!

Issue #1019, opened 2024-05-14
Flickering on Windows when using bubbletea >=0.26.0


supleed2:
**Describe the bug**
I tried updating to Bubbletea v0.26.0 (and also .1/.2) but the output seems to flicker a lot when running on Windows.

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS: Windows 11 Build 22621
 - Shell: Powershell Core (also tried on cmd)
 - Terminal Emulator: Windows Terminal (also tried on Wezterm)
 - Terminal Multiplexer: None

**To Reproduce**
I don't have a minimal reproducible example yet, but noticed this while updating dependencies on my project, linked below. Specifically the client binary which uses bubbletea for the UI.

**Source Code**
Available here: [supleed2/go-chat](https://github.com/supleed2/go-chat/blob/main/client/main.go)

**Expected behavior**
Stable output as with v0.25.0, observed with binaries available at: https://github.com/supleed2/go-chat/releases/tag/v0.1.2

Ideally with the new ability to resize :)

**Screenshots**
GIF to demonstrate:

![clip](https://github.com/charmbracelet/bubbletea/assets/21363892/066bf83c-96d9-41aa-ba33-3c98dd545e24)

meowgorithm:
Okay awesome, thanks for reporting this. There's seems to be a bug in Windows where going in and out of the altscreen can fire off an infinite loop of `WindowResizeMsg`s. This wasn't the case in prior versions of Bubble Tea because resize events weren't possible.

We're looking into a fix but for your immediate case you can actually just remove [this line](https://github.com/supleed2/go-chat/blob/36047b07d124c09b9a322606ab3b86235f8ed6a8/client/main.go#L249), which _seems_ to have been left in by mistake:

```diff
diff --git a/client/main.go b/client/main.go
index b8fe61a..37b8bd7 100644
--- a/client/main.go
+++ b/client/main.go
@@ -246,7 +246,6 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			m.input.Reset()
 		}
 	case tea.WindowSizeMsg:
-		clCmd = tea.Sequence(tea.ExitAltScreen, tea.EnterAltScreen)
 		m.history.Height = msg.Height - 2
 		m.history.Width = msg.Width
 		m.history.GotoBottom()
```

supleed2:
🤦 I just have to say thank you so much for digging through and noticing what was wrong with *my* code.

[...]

PR #1021, opened 2024-05-16
Prevent multiple window-size-events from firing on Windows

aymanbagabas:
Windows alt-screen-buffer changes trigger window-size-events causing Bubble Tea to send a new event even if the window size is the same as before. This simply checks if the window size is the same before bubbling it to the model.

Fixes: https://github.com/charmbracelet/bubbletea/issues/1019

### DIFF
--- a/inputreader_windows.go
+++ b/inputreader_windows.go
@@ -17,16 +17,9 @@ import (
 type conInputReader struct {
 	cancelMixin
 
-	conin       windows.Handle
-	cancelEvent windows.Handle
+	conin windows.Handle
 
 	originalMode uint32
-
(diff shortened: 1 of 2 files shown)


meowgorithm:
@aymanbagabas will this still send `WindowSizeMsg`s when going in and out of the alt screen?

caarlos0 (approved):

aymanbagabas:
> @aymanbagabas will this still send `WindowSizeMsg`s when going in and out of the alt screen?

This won't send more than one `WindowSizeMsg` with the same size, so yes it will still send the msg but _only if_ the size changes.

meowgorithm (approved):

PR #1022, opened 2024-05-22
fix: prefer `tea.Tick` over goroutines

jshawl:
This pull request replaces the goroutine with `tea.Tick()` in the send-msg example.

From the blog post [Commands in Bubble Tea](https://charm.sh/blog/commands-in-bubbletea/):

> Never use goroutines within a Bubble Tea program

I used this example to implement sending messages at an interval, but later discovered the tip in the blog post. Hopefully this example will communicate the best practice in addition to the blog post.

Thanks!

### DIFF
--- a/examples/send-msg/main.go
+++ b/examples/send-msg/main.go
@@ -1,7 +1,7 @@
 package main
 
-// A simple example that shows how to send messages to a Bubble Tea program
-// from outside the program using Program.Send(Msg).
+// A simple example that shows how to send messages at an interval
+// using tea.Tick()
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


examples/send-msg/main.go
```diff
@@ -52,8 +52,17 @@ func newModel() model {
 	}
 }
 
+func doTick() tea.Cmd {
```
jshawl, line 55: reusing the function name from the docs - https://pkg.go.dev/github.com/charmbracelet/bubbletea#Tick

examples/send-msg/main.go
```diff
@@ -52,8 +52,17 @@ func newModel() model {
...
+	// Simulate activity
+	pause := time.Duration(rand.Int63n(899)+100) * time.Millisecond // nolint:gosec
+	return tea.Tick(pause, func(t time.Time) tea.Msg {
+		// Send a resultMsg to the Update method
```
jshawl, line 59: I don't think other models in a single `tea.NewProgram` will receive this message, but if so it might be good to clarify that behavior here, e.g. not just this `Update` method

meowgorithm:
Thanks for the PR. So in this particular example, we're illustrating how to inject messages from outside a program with `Program.Send()`. You’re correct that Goroutines shouldn't be used within a program, however in this case the Goroutine is actually outside the program in `main`.

So all said, your changes are good ones in terms of how we generally recommend writing Bubble Tea programs, but in this case I'd prefer if we left this example as-is to illustrate how `Program.Send` works.

jshawl:
Thank you @meowgorithm! Hopefully I’ll be back with another contribution in the future :)

Issue #850, opened 2023-11-01
Support for Synchronized Output


talios:
**Is your feature request related to a problem? Please describe.**

Recently I've been beta testing the new Ghostty terminal (https://mitchellh.com/ghostty), whose super fast performance has started revealing various refresh/tearing issues in other applications.

When running the Dagger (dagger.io) CI tool locally, and more so using experimental UI, I noticed A LOT of screen flicker, after discussing this on both the Ghostty and Dagger discords, I was pointed here.

Ghostty supports the Synchronized Output spec (https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036, inspired from iTerm) which aims to provide a contract between terminals and applications to reduce screen tearing and apparent UI flickering.

**Describe the solution you'd like**

I'm not a bubbletea user myself, but a user of tools _using_ bubbletea, but I'd love to see the library adopt support for Synchronised Output and makes many users a lot less... tearful.

meowgorithm:
Hi! Thanks for filing; this makes sense to me. No ETA on when we can commit to it, though at first read it looks like implementation would be pretty straightforward.

We're actually also Ghostty users, though I personally haven't seen much jank yet. I will say that traditionally you'll see some lag when redrawing very large TUIs on very large windows, and we do plan to address that in other ways as well.

Few more questions, which will help us measure results a bit:

1. Are you using very large terminal?
2. Any chance you upload a screencast of the behavior?
3. Which OS are you using? 
5. Are you seeing this with any other tools built on Bubble Tea? Asking mostly because at first glance Dagger looks like it may take a degree of setup before we can actually see some UI from it (I could be wrong through).

talios:
Hi, just a quick response and its 3:30am (I couldn't sleep).

This is a link to my ghostty discord message/video: https://discord.com/channels/1005603569187160125/1005603569711452192/1168783411968745482 - I can post that elsewhere in the morning if not accessible.

The terminal window isn’t that large even even small term sizes produce the effect.

This is on macOS Sonoma (Intel).

muesli:
I do have a patch implementing synchronized output. I've never followed up on it, as so far I had not seen a terminal where it would make a difference, but if ghostty benefits from this, I'll happily open a PR.

talios:
@meowgorithm the bottom version was WezTerm, with Ghostty at the top - which as Mitchel pointed out, still shows some tearing but not as much as Ghostty - mostly cause Ghostty is much more responsive.

[...]

PR #1027, opened 2024-05-29
feat: Synchronized output

nervo:
This is an attempt to implement synchronized output in bubbletea. At this time, this is so naive, that you can see it as a proof-of-concept :)

Anyway, this does not break on my machine 👍 

Some interesting reading:
- https://gitlab.com/gnachman/iterm2/-/wikis/synchronized-updates-spec
- https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036
- https://hugotunius.se/2019/12/29/efficient-terminal-drawing-in-rust.html#going-faster
- https://github.com/charmbracelet/bubbletea/issues/850

Feel free to comment ❤️

### DIFF
--- a/standard_renderer.go
+++ b/standard_renderer.go
@@ -47,6 +47,9 @@ type standardRenderer struct {
 	// whether or not we're currently using bracketed paste
 	bpActive bool
 
+	// whether or not we're currently using synchronized output
+	soActive bool
+
 	// renderer dimensions; usually the size of the window
 	width  int
 	height int
@@ -274,7 +277,18 @@ func (r *standardRenderer) flush() {
 		buf.WriteString(ansi.CursorLeft(r.width))
 	}
 
+	// Enable synchronized output
+	if r.soActive {
+		_, _ = io.WriteString(r.out, ansi.EnableSyncdOutput)
+	}
+
 	_, _ = r.out.Write(buf.Bytes())
+
+	// Disable synchronized output
+	if r.soActive {
+		_, _ = io.WriteString(r.out, ansi.DisableSyncdOutput)
+	}
+
 	r.lastRender = r.buf.String()
 	r.buf.Reset()
 }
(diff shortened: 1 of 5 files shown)


vito:
Friendly poke to whoever can keep this moving, could really use this for my TUI as it flickers pretty heavily on certain terminals. Cool to see a PR up for it already! 😄

nervo:
@vito do you think you can give a try ?

Dear maintainners, are you ok with the way the feature is implemented (wording of methods and variables) ?

For the feature itself, i'm not an ansi guru, i just need to know:
1. if i made the terminal query the good way
2. If i understood the protocol correctly (the `event.Value` 2 or 4)
3. if i'm doing the right things: sending the enable code just before, and the disable code right after writing the buffer, both  independently from the buffer itself.

vito (commented):
Tried it out and went looking for how to enable it, and noticed it just enables itself if the terminal supports it, which makes sense to me. :+1: 

I think I found a typo along the way though - left a suggestion.

standard_renderer.go
```diff
@@ -459,6 +473,13 @@ func (r *standardRenderer) setWindowTitle(title string) {
 	r.execute(ansi.SetWindowTitle(title))
 }
 
+func (r *standardRenderer) enableSynchronizedOutput() {
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+
+	r.bpActive = true
```
vito, line 480: Err, my dog ate my suggestion, here it is:

```suggestion
	r.soActive = true
```

nervo: Oh god 😱 
Thx a lot @vito  !

meowgorithm:
Note: this will close #850

meowgorithm:
Hey @nervo! Thanks for looking into this one, it's something we've very much been wanting to get in.

API aside, the big thing that will need to happen for this feature to work is an implementation at the renderer level so that Bubble Tea can repaint in-step with the terminal. We're currently working on an overhaul to the renderer, so we should hold off on this until we've made that update.

nervo:
@meowgorithm ok, good to hear 👍

PR #1028, opened 2024-05-29
fix: wait for goland terminal size

nervo:
The embedded terminal of JetBrains products (JediTerm) such as Goland have a serious bug :(

At the startup, It waits for approx. 400ms before answering *real* window sizes... As a matter of consequence, in many situations, the first `WindowSizeMsg` received by models is width=0 and height=0

This pr tries to mitigate this issue by repeating nth requests during an arbitrary delay of 600ms on the first call.

- Fix https://github.com/charmbracelet/huh/issues/204
- Address https://github.com/charmbracelet/huh/discussions/157

Please, let me know if you know or find any other link referencing this issue ❤️

### DIFF
--- a/signals_unix.go
+++ b/signals_unix.go
@@ -28,6 +28,6 @@ func (p *Program) listenForResize(done chan struct{}) {
 		case <-sig:
 		}
 
-		p.checkResize()
+		p.checkResize(false)
 	}
 }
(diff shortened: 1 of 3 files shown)


meowgorithm:
Hey, @nervo! Oof. No opinions on this one yet, but under the hood this may be able to make use of the new `WindowSize` internally command introduced last week to reduce code duplication: https://github.com/charmbracelet/bubbletea/pull/988.

denizgursoy:
Is this going to be merged?

PR #1046, opened 2024-06-28
fix(examples): Fix bugs in package-manager

AkshayKalose:
I fixed 3 bugs in the package-manager example (they are easier to spot when working with a small list of 4, for example):

1. The total packages count.
2. The progress bar percentage.
3. The final installed package in the output.

## Before
before

## After
after


meowgorithm (approved):
_Amazing_! Thanks so much for the fix!

PR #1054, opened 2024-07-11
feat: allow to suspend bubbletea programs

### LINKED ISSUE #497: bug: cannot suspend bubbletea process

### LINKED ISSUE #1053: Suspend/Resume
It would be nice to be able to support suspend and resume (aka `^z` and `fg`). I imagine we'd need to listen for the appropriate signals as well as implement `Cmd`s.

Related: https://github.com/charmbracelet/glow/issues/333

caarlos0:
closes #497
closes #1053

based on @knz's implementation. thanks ❤️

### DIFF
--- a/examples/altscreen-toggle/main.go
+++ b/examples/altscreen-toggle/main.go
@@ -14,8 +14,9 @@ var (
 )
 
 type model struct {
-	altscreen bool
-	quitting  bool
+	altscreen  bool
+	quitting   bool
+	suspending bool
 }
 
 func (m model) Init() tea.Cmd {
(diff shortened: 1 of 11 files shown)


knz:
Thank you!

knz (commented):
Typo in docstring

caarlos0:
also made the program block and wait for a `SIGCONT` after sending `SIGTSTP` ⏲️

meowgorithm (approved):
This is _so_ good. @knz, thanks for sharing your solve for this!

PR #1063, opened 2024-07-18
feat: set the program environment variables

aymanbagabas:
This adds the option to provide a custom set of environment variables to use in the Bubble Tea program. When running Bubble Tea in a remote session such as SSH, we need to be able to pass the SSH session's environment variables rather than the server ones.

```go
	// Add the TERM from the SSH session to the environment
	var sess ssh.Session // from the github.com/charmbracelet/ssh package
	pty, _, _ := sess.Pty()
	environ := append(sess.Environ(), "TERM="+pty.Term)
	p := tea.NewProgram(model{}, tea.WithEnvironmen(environ))
```

Now, the Bubble Tea program running for a remote session can determine the appropriate color profile from the session's environment variables instead of the _server's_ environment variables.

In most cases you'll only need this feature when using Bubble Tea in remote sessions.

### DIFF
--- a/options.go
+++ b/options.go
@@ -49,6 +49,23 @@ func WithInputTTY() ProgramOption {
 	}
 }
 
+// WithEnvironment sets the environment variables that the program will use.
+// This useful when the program is running in a remote session (e.g. SSH) and
+// you want to pass the environment variables from the remote session to the
+// program.
+//
+// Example:
+//
+//	var sess ssh.Session // ssh.Session is a type from the github.com/charmbracelet/ssh package
+//	pty, _, _ := sess.Pty()
+//	environ := append(sess.Environ(), "TERM="+pty.Term)
+//	p := tea.NewProgram(model, tea.WithEnvironment(environ)
+func WithEnvironment(env []string) ProgramOption {
+	return func(p *Program) {
+		p.environ = env
+	}
+}
+
 // WithoutSignalHandler disables the signal handler that Bubble Tea sets up for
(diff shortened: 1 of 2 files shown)


meowgorithm (approved):
This is awesome

options.go
```diff
@@ -49,6 +49,13 @@ func WithInputTTY() ProgramOption {
 	}
 }
 
+// WithEnvironment sets the environment variables that the program will use.
```
meowgorithm, line 52: I'd elaborate here on the exact use cases.

meowgorithm:
Two questions with this one:

* Will it have any effect on local programs?
* How would the user implement it in Wish?

I'd also add a very short example to this PR (and probably the doc comments) so we can reference it later.

aymanbagabas:
Added an example in the code comment

> * Will it have any effect on local programs?

No, bubbletea programs default to `os.Environ()` if no environment is set.

> * How would the user implement it in Wish?

The bubbletea middleware should pass the SSH session's environment to the program. The middleware should include something like:

```go
	var sess ssh.Session // ssh.Session is a type from the github.com/charmbracelet/ssh package
	pty, _, _ := sess.Pty()
	environ := append(sess.Environ(), "TERM="+pty.Term)
	p := tea.NewProgram(model, tea.WithEnvironment(environ)
```

caarlos0 (approved):

meowgorithm (approved):

PR #1064, opened 2024-07-18
Windows: ensure output is UTF16 encoded when output is not a terminal

aymanbagabas:
We need to transform the output from UTF8 to wide characters UTF16 for Windows to understand when output is not a terminal.


caarlos0 (approved):

aymanbagabas:
We don't need this anymore, closing.

PR #1079, opened 2024-08-12
feat: support focus/blur csi sequences

caarlos0:


caarlos0:
this is in main already

Issue #1014, opened 2024-05-10
Proposal: use x/input to handle input events


aymanbagabas:
Current Bubble Tea input handling is quite limiting as it doesn't support parsing input events such as cursor position report, fixterms & Kitty keyboard, and other terminal features. This is because we only do map lookups for common key sequences used by most terminals.

 This is where the work on `charmbracelet/x/input` began. This new `input` package can replace the existing Bubble Tea input handler to enhance and improve parsing input events and support. This means we can support more terminal features since we can now parse input sequences such as the ones found in the Kitty keyboard protocol and others.

However, this comes with some breaking changes since the `input` package defines keys, mouse, and other events differently. For instance, `KeyMsg` is now split into 2 different events, a `KeyUpEvent` and a `KeyDownEvent`. With this, we can distinguish between a key press and a key release. The same goes for mouse events, we can detect `MouseDownEvent`, `MouseUpEvent`, `MouseWheelEvent`, and `MouseMotionEvent`.

Another example is how `input` handles bracketed-paste events. Using `input`, bracketed-paste events are their own event type `input.PasteEvent` which includes the string that was pasted along with two empty events `input.PasteStartEvent` and `input.PasteEndEvent` indicating the start and end of the bracketed-paste event.

This means that the current API, especially for `KeyMsg` and `MouseMsg`, will change and `KeyType`s like `KeyCtrlA` will get replaced with `input.KeySym` and `input.KeyMod` representing a keyboard key symbol and a keyboard modifier respectfully.

With this being said, the proposed API is to simply type alias these changes and direct users to use `input.KeyEvent.String()` to do the key matching.

```go
// KeyMsg represents a keyboard key press/release msg.
type (
	Key        = input.Key
	KeyMsg     = input.KeyDownEvent
	KeyDownMsg = input.KeyDownEvent
	KeyUpMsg   = input.KeyUpEvent
)

// Modifier keys.
const (
	Shift = input.Shift
	Alt   = input.Alt
	Ctrl  = input.Ctrl
	Meta  = input.Meta

	// These modifiers are used with the Kitty protocol.

	Hyper = input.Hyper
	Super = input.Super // Windows/Command keys

	// These are key lock states.

	CapsLock   = input.CapsLock
	NumLock    = input.NumLock
	ScrollLock = input.ScrollLock // Defined in Windows API only
)

// Symbol constants.
const (
	KeyNone = input.KeyNone

	// Special names in C0

[...]

Issue #918, opened 2024-02-06
Add support for setting terminal cursor position


indiejames:
**Is your feature request related to a problem? Please describe.**
I am unable to set the terminal cursor position reliably (the real one, not the one rendered by the TextArea bubble). I have tried using direct ANSI escape codes as well as instantiating a separate `termenv` `Output` structure. In both cases the cursor can be moved, but as soon as I render anything with my UI method the cursor pops back down to the bottom left corner of the terminal. 

Note that I am able to make the cursor visible by returning a `tea.ShowCursor` message from my `Update` method.

**Describe the solution you'd like**
I would like to see a new command generator function added to the API, similar to the `SetWindowTitle` function. Perhaps
```
func SetCursorPosition(row int, column int) Cmd
```
Using this function should set the position of the cursor and it should stay in that position until the method is called again.

**Describe alternatives you've considered**
I know that the `TextArea` bubble uses a rendered cursor by setting a background color on the character at the cursor position. This works for a block cursor, and an underscore cursor could work similarly, but I'm not aware of a way to render a bar cursor (vertical line to the left of the character under the cursor) with this approach.

**Additional context**

muesli:
I don't think this is something we'll support, sorry. For the tea renderer to work, it needs to reset the cursor position before rendering.

meowgorithm:
Contextual note here that we should ultimately support this for IME (see https://github.com/charmbracelet/bubbletea/issues/874) — however we'll need to reexamine the renderer to get there and I don't see that happening in the short term.

indiejames:
Thanks for the feedback. I have moved on to tview/tcell for my application, although I still love bubbletea and would use it if I could control the cursor. But I understand that this is not a priority for most users. Feel free to close this issue or leave it open as you see fit.

realtimeprojects:
I agree... not having a vertical line as a cursor in a 2025 terminal ui's is really outdated. This is the issue killing my usage of this library as well. You don't even need to implement placing the cursor randomly.

aymanbagabas:
> I agree... not having a vertical line as a cursor in a 2025 terminal ui's is really outdated. This is the issue killing my usage of this library as well. You don't even need to implement placing the cursor randomly.

This is possible with Bubble Tea v2. Take a look at https://github.com/charmbracelet/bubbletea/discussions/1374

Issue #869, opened 2023-11-14
Kitty Keyboard Protocol


Broderick-Westrope:
**Is your feature request related to a problem? Please describe.**
It would be nice to see a solution to detecting key release events. Currently, for any action that requires this functionality in my Bubbletea apps I have replaced it with pressing the key to toggle which is less than ideal.

**Describe the solution you'd like**
The ideal solution would be to have an `actionType` (or similar) field that determines what happened (press, release, etc). Unfortunately, this would likely break a lot of existing code, so another solution would be to add release versions of all the basic keys.

**Describe alternatives you've considered**
I've considered cloning the Bubbletea package for projects that require it, and adding the functionality using solutions such as [gohook](https://github.com/robotn/gohook). I am unsure of other potential solutions.

**Additional context**
If anyone has other ideas of how this could be achieved, either on my own or how I could implement this better into the Bubbletea package, please let me know!

meowgorithm:
Hi! So if you're targeting all terminals you simply cannot detect release events (terminal input handing is rather archaic).

We can, however, offer support release events, when available, with the [Kitty keyboard protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#fixterms-bugs). Keep in mind, however, that there are only a few terminals that support the protocol at the moment.

## Maintainer's notes:

fixterms/`CSI u` (https://github.com/charmbracelet/bubbletea/issues/293), also proposed, does not support key release events. The Kitty keyboard is mostly backwards compatible with this protocol, but not entirely.

A large, high fidelity range of input options, including key release, are available in Windows as evidenced in https://github.com/charmbracelet/bubbletea/pull/140. We didn't merge that because of the large amount of Windows-only code and the maintenance burden that would come with it.

meowgorithm:
Just a note that this is currently in development on the [`term-input`](https://github.com/charmbracelet/bubbletea/tree/term-input) branch.

Broderick-Westrope:
Hey @meowgorithm, wondering if there is any update on this? I notice that the branch went untouched very quickly (March 1st). Has the work moved elsewhere or is there a plan for future implementation of the Kitty keyboard protocol? Thanks

ajayd-san:
Same, I'm also waiting for this to be implemented. I'd like to take the issue up if no one is working on it

meowgorithm:
[...]

PR #1080, opened 2024-08-12
refactor: reimplement the event loop with a sequence parser

aymanbagabas:
Currently, Bubble Tea uses a simple lookup table to detect input events. Here, we're introducing an actual input sequence parser instead of simply using a lookup table. This will allow Bubble Tea programs to read all sorts of input events such Kitty keyboard, background color, mode report, and all sorts of ANSI sequence input events.

This PR includes the following changes:
- Support clipboard OSC52 read messages (`OSC 52 ?`)
- Support terminal foreground/background/cursor color report messages (OSC10, OSC11, OSC12)
- Support terminal focus events (mode 1004)
- Deprecate the old `KeyMsg` API in favor of `KeyPressMsg` and `KeyReleaseMsg`
  - `KeyType` const values are different now. Programs that use int value comparison **will** break. E.g. `key.Type == 13` where `13` is the control code for `CR` that corresponds to the <kbd>enter</kbd> key. (BREAKING CHANGE!)
  - Bubble Tea will send two messages for key presses, the first of type `KeyMsg` and the second of type `KeyPressMsg`. This is to keep backwards compatibility and _not_ break the API
  - `tea.Key` contains breaking changes (BREAKING CHANGE!)
- Deprecate `MouseMsg` in favor of `MouseClickMsg`, `MouseReleaseMsg`, `MouseWheelMsg`, and `MouseMotionMsg`
  - Bubble Tea will send two messages for mouse clicks, releases, wheel, and motion. The first message will be a `MouseMsg` type. And the second will have the new corresponding type. This is to keep backwards compatibility and _not_ break the API
  - `tea.Mouse` contains breaking changes (BREAKING CHANGE!)
- Support reading Kitty keyboard reports (reading the results of sending `CSI ? u` to the terminal)
- Support reading Kitty keyboard and fixterms keys `CSI u`
- Support reading terminal mode reports (DECRPM)
- Bracketed-paste messages now have their own message type `PasteMsg`. Use `PasteStartMsg` and `PasteEndMsg` to listen to the start/end of the paste message.
  - Bubble Tea will send two messages for bracketed-paste, the first is of type `KeyMsg` and the second is of type `PasteMsg`. This is to keep backwards compatibility and _not_ break the API
- Support more obscure key input sequences found in URxvt and others
- Support reading termcap/terminfo capabilities through `XTGETTCAP`. These capabilities will get reported as `TermcapMsg`
- Support reading terminfo databases for key input sequences (disabled for now)
- Support reading [Win32 Input Mode keys](https://github.com/microsoft/terminal/blob/main/doc/specs/%234999%20-%20Improved%20keyboard%20handling%20in%20Conpty.md#win32-input-mode-sequences)
- Support reading Xterm `modifyOtherKeys` keys

TODO:
- [x] Parse multi-rune graphemes as one `KeyPressMsg` storing it in `key.Runes`
- [x] Kitty keyboard startup settings and options https://github.com/charmbracelet/bubbletea/pull/1083
- [x] Xterm modify other keys startup settings and options https://github.com/charmbracelet/bubbletea/pull/1084
- [x] Focus events startup settings and options https://github.com/charmbracelet/bubbletea/pull/1081
- [x] Fg/bg/cursor terminal color startup settings and options https://github.com/charmbracelet/bubbletea/pull/1085

Supersedes: https://github.com/charmbracelet/bubbletea/pull/1079
Supersedes: https://github.com/charmbracelet/bubbletea/issues/1014
Related: https://github.com/charmbracelet/bubbletea/issues/869
Related: https://github.com/charmbracelet/bubbletea/issues/163
Related: https://github.com/charmbracelet/bubbletea/issues/918
Related: https://github.com/charmbracelet/bubbletea/issues/850
Related: https://github.com/charmbracelet/bubbletea/issues/207


caarlos0 (approved):

renderer.go
```diff
@@ -36,40 +36,12 @@ type renderer interface {
 	// Hide the cursor.
 	hideCursor()
 
-	// enableMouseCellMotion enables mouse click, release, wheel and motion
-	// events if a mouse button is pressed (i.e., drag events).
-	enableMouseCellMotion()
```
caarlos0, line 41: one key difference in all these methods is that they were locking the renderer before writing... not sure if we need that, but i thought it was worth mentioning anyway

aymanbagabas: It won't matter because the renderer eventually writes to the program output, _not_ the renderer buffer

caarlos0 (approved):

parse_test.go
```diff
@@ -0,0 +1,41 @@
+package tea
+
+import (
+	"image/color"
+	"reflect"
+	"testing"
+)
+
+func TestParseSequence_Events(t *testing.T) {
+	input := []byte("\x1b\x1b[Ztest\x00\x1b]10;rgb:1234/1234/1234\x07\x1b[27;2;27~\x1b[?1049;2$y")
```
bashbunni, line 10: should we maybe be generating the input here programmatically? If this test fails, it seems like it might be a bit of a headache to figure out what went wrong with this sequence

aymanbagabas: Always good to have more tests. We already do for key sequences in `key_test.go`.

aymanbagabas: I've added more tests for Kitty/fixterms, Xterm modifyOtherKeys, and terminal bg/fg/cur colors in their respective PRs.

PR #1085, opened 2024-08-14
feat: query and set terminal background, foreground, and cursor colors

aymanbagabas:
This adds the ability to read and set terminal background, foreground, and cursor color.

Supersedes: https://github.com/charmbracelet/bubbletea/pull/567
Fixes: https://github.com/charmbracelet/bubbletea/issues/207

### DIFF
--- a/color.go
+++ b/color.go
@@ -3,9 +3,64 @@ package tea
 import (
 	"fmt"
 	"image/color"
+	"math"
 	"strconv"
 	"strings"
 )
+
+// backgroundColorMsg is a message that requests the terminal background color.
+type backgroundColorMsg struct{}
+
+// BackgroundColor is a command that requests the terminal background color.
+func BackgroundColor() Msg {
+	return backgroundColorMsg{}
+}
+
+// foregroundColorMsg is a message that requests the terminal foreground color.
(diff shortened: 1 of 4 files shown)


caarlos0 (approved):

color.go
```diff
@@ -80,3 +150,68 @@ func xParseColor(s string) color.Color {
...
+	} else {
+		max = b
+		min = a
+	}
+	if c > max {
+		max = c
+	} else if c < min {
+		min = c
+	}
+	return max, min
```
caarlos0, line 167:
```suggestion
	return max(a, b, c), min(a, b, c)
```

aymanbagabas: `min` and `max` require go1.21

caarlos0: ah, right, could use `ordered.Min` and `ordered.Max` then... 

https://github.com/charmbracelet/x/blob/main/exp/ordered/ordered.go

I think it would look clearer

aymanbagabas: I don't want to introduce more dependencies. I will put a todo note for now until we bump the go version

caarlos0: fair

Issue #982, opened 2024-04-19
Async Clipboard


meowgorithm:
I've been seeing reports that clipboard operations can take a long time in WSL. While moving copy/paste to `Cmd`s won't solve this issue, it's probably worth doing it regardless, both because clipboard stuff is I/O, and to provide a unified copy/paste API to all Bubble Tea applications.

Note that we also discussed this in @aymanbagabas's upcoming Kitty Keyboard work.

ajayd-san:
Hey can I take this up ?

meowgorithm:
Sounds good to me. @aymanbagabas, whatcha think?

aymanbagabas:
I think that would be wonderful to add to Bubble Tea!

Looking at implementation details, there are a couple of golang clipboard libraries such as https://github.com/atotto/clipboard and https://github.com/golang-design/clipboard, but each has its own shortcomings. The former always assumes programs running locally (no remote sessions support) and the latter uses CGO.

Given this, I'm proposing porting atotto/clipboard to bubbletea with a few changes to recognize the [program environment variables](https://github.com/charmbracelet/bubbletea/pull/1063) and use [OSC52](https://github.com/charmbracelet/x/blob/main/ansi/clipboard.go#L20) when working with remote sessions like SSH. Something along the lines of this API

```go
type ClipboardWriteMsg bool

func ClipboardWrite(text string) tea.Cmd {
  return func() tea.Msg {
    err := copyClipboard()
    return ClipboardWriteMsg(err == nil)
  }
}

type ClipboardReadMsg string

func ClipboardRead() tea.Cmd {
  return func() tea.Msg {
    text := readClipboard()
    return ClipboardReadMsg(text)
  }
}
  
```

ajayd-san:
Hey, So tell me if I'm going in the right direction. 

I'm thinking of adding `ClipboardRead` and `ClipboardWrite` functions in `commands.go` and I'll be using `atotto/clipboard`  for normal sessions and `x/clipboard` for ssh session. 

How do i know if I'm in a ssh session? Check if `SSH_CLIENT` or `SSH_TTY` is set in `p.environ`. 

Does this look good?

aymanbagabas:
> I'll be using `atotto/clipboard` for normal sessions

That wouldn't work because `atotto/clipboard` uses `os.Getenv()` instead of `p.environ` to determine environment variables. We need to port `atotto/clipboard` to bubbletea and make it use the program's environment variables.

[...]

PR #1086, opened 2024-08-15
feat: support setting and querying the terminal clipboard using OSC52

aymanbagabas:
This adds support to setting the system and primary (X11 & Wayland) clipboards using OSC52. This makes the clipboard commands work even on remote session such as SSH.

While this doesn't work on all terminals, most modern terminals support OSC52 including Alacritty, Kitty, Xterm.JS, etc. For terminals, that don't support OSC52, application developers should consider using a Golang clipboard library like https://github.com/atotto/clipboard.

OSC52 support can be detected if the terminal responds to a `ReadClipboard` command.

Fixes: https://github.com/charmbracelet/bubbletea/issues/982

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -581,15 +581,28 @@ func parseOsc(b []byte) (int, Msg) {
 		if len(parts) == 0 {
 			return i, ClipboardMsg("")
 		}
-		b64 := parts[len(parts)-1]
-		bts, err := base64.StdEncoding.DecodeString(b64)
-		if err != nil {
-			return i, ClipboardMsg("")
+		if len(parts) != 2 {
+			break
 		}
-		return i, ClipboardMsg(bts)
-	default:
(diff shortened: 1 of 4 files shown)


parse.go
```diff
@@ -581,15 +581,28 @@ func parseOsc(b []byte) (int, Msg) {
 		if len(parts) == 0 {
 			return i, ClipboardMsg("")
 		}
-		b64 := parts[len(parts)-1]
+		if len(parts) != 2 {
```
caarlos0, line 584: maybe use `strings.Cut`? not sure if it existed in 1.18 or only earlier version

caarlos0 (approved):

parse.go
aymanbagabas: But we also want to ensure that we exactly have 2 parts, `strings.Cut` will only split on the first occurrence of `;`.

PR #1118, opened 2024-08-28
V2 exp

aymanbagabas:


caarlos0:
these do not work on freebsd: https://github.com/charmbracelet/bubbletea/blob/12f6445fb2c4aad680359f14687ac636f7516099/tty_unix.go#L27-L28

tty_unix.go
```diff
@@ -20,9 +21,14 @@ func (p *Program) initInput() (err error) {
 		if err != nil {
 			return fmt.Errorf("error entering raw mode: %w", err)
 		}
+
+		// OPTIM: We can use hard tabs to optimize cursor movements if the
+		// terminal doesn't have tab expansion enabled.
+		p.useHardTabs = p.previousTtyInputState.Oflag&unix.TABDLY == unix.TAB0
+		p.useBackspace = p.previousTtyInputState.Lflag&unix.BSDLY == unix.BS0
```
caarlos0, line 28: this does not work in freebsd, openbsd, dragonfly, not sure about zos

aymanbagabas: Hmm let me take a look. Might need to make an exception for those

agca52 (approved):
👍

Issue #1069, opened 2024-07-29
tui-daemon-combo example output is not aligned correctly in daemon mode


ptxmac:
**Describe the bug**
Running the example `tui-daemon-combo` in daemon mode (`-d`) renders the output without positioning each new line correctly.
See attached recording:
[![asciicast](https://asciinema.org/a/cvmUEUj45EcxgSRVqOkt6j406.svg)](https://asciinema.org/a/cvmUEUj45EcxgSRVqOkt6j406)

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS: MacOS & Linux
 - Shell: fish and bash
 - Terminal Emulator: iterm2

**To Reproduce**
1. `git clone git@github.com:charmbracelet/bubbletea.git`
2. `cd bubbletea/examples/tui-daemon-comb`
3. ˚`go run . -d`
4. :tada: Broken lines

ptxmac:
It looks like this is because the input terminal is always converted to raw, even if the output isn't a terminal / the render is the nil-renderer: https://github.com/charmbracelet/bubbletea/blob/master/tty_unix.go#L19

PR #1120, opened 2024-08-28
fix: we don't initialize the terminal when using a nilRenderer

aymanbagabas:
Otherwise, a raw terminal will mess up the output. This is because a raw terminal disables termios OPOST mode which converts newlines to `\r\n` to reset the cursor to the beginning of the screen on new lines.

CC/ @cwarden

Fixes: https://github.com/charmbracelet/bubbletea/issues/1069

### DIFF
--- a/tty.go
+++ b/tty.go
@@ -23,6 +23,11 @@ func (p *Program) suspend() {
 }
 
 func (p *Program) initTerminal() error {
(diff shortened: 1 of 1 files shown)


meowgorithm (approved):
Niiiiice

PR #1121, opened 2024-08-29
feat: use go 1.21

caarlos0:
- some dependency is already requiring it, as examples/go.mod was set to 1.21 already 
- builds were breaking because of it
- go 1.18 is really old, maybe its time to upgrade?


caarlos0:
ah, just saw 105d88a... yeah maybe that's ought to be enough

aymanbagabas (approved):

PR #1130, opened 2024-09-04
docs(featured): clean up in the wild list

bashbunni:
Preview: https://github.com/charmbracelet/bubbletea/blob/in-the-wild/README.md#bubble-tea-in-the-wild


meowgorithm (approved):
I think we’re good to go here for now if you are @bashbunni (if not, edit away). There are a few more apps that would be good to feature but we can dig 'em up later.

bashbunni:
Ok sounds good, I'll merge in a bit

Issue #1032, opened 2024-06-01
When a large update is performed, it will cause a flash.


yorukot:
**Describe the bug**
When a large update is performed, it will cause a flash.

**Setup**
 - OS: Fedora
 - Shell: zsh
 - Terminal Emulator: kitty
 - Terminal Multiplexer: zellij
 
**To Reproduce**
1. go run main.go
2. Press Tab to let tui update

**Source Code**
```golang
package main

import (
	tea "github.com/charmbracelet/bubbletea"
	"github.com/charmbracelet/lipgloss"
	"log"
	"math/rand"
	"strings"
)

type sessionState uint

const (
	timerView sessionState = iota
	spinnerView
)

var (
	modelStyle = lipgloss.NewStyle().
			Width(80).
			Height(35).
			Align(lipgloss.Center, lipgloss.Center).
			BorderStyle(lipgloss.HiddenBorder()).
			Background(lipgloss.Color("#4B4B4B"))

	focusedModelStyle = lipgloss.NewStyle().
				Width(80).
				Height(35).
				Align(lipgloss.Center, lipgloss.Center).
				BorderStyle(lipgloss.NormalBorder()).
				BorderForeground(lipgloss.Color("69")).
				Background(lipgloss.Color("#4B4B4B"))
)

type mainModel struct {
	state sessionState
}

func newModel() mainModel {
	m := mainModel{}
	return m
}

func (m mainModel) Init() tea.Cmd {
	return tea.Batch()
}

func (m mainModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	var cmds []tea.Cmd
	switch msg := msg.(type) {
	case tea.KeyMsg:
		switch msg.String() {
		case "ctrl+c", "q":
			return m, tea.Quit
		case "tab":
			if m.state == timerView {
				m.state = spinnerView
			} else {
				m.state = timerView
			}
		}
	}
	return m, tea.Batch(cmds...)
}

func (m mainModel) View() string {
	var s string
	example1 := strings.Repeat("AAAA", 760)
	example2 := strings.Repeat("Test", 760)
	example := []string{example1, example2}
	if m.state == timerView {
		s += lipgloss.JoinHorizontal(lipgloss.Top, focusedModelStyle.Render(example[rand.Intn(2)]), modelStyle.Render(example[rand.Intn(2)]))
	} else {
[...]

PR #1132, opened 2024-09-08
feat(render): improve renderer; remove flickering

LeperGnome:
Hey! This is my naive approach to fixing output flickering. Here are my thoughts behind the solution:
- Flickering is caused by `ansi.EraseEntireLine` escape chars, because it allows the time, when there's an empty output.
- So instead we just move cursor to the beginning of the drawing section on each new frame and write over latest frame, thus replacing old characters.
- Erase characters from the old frame on each line end with `ansi.EraseLineRight`
- Erase lines from old frame with `ansi.EraseDisplayRight`

I tested #1032 and it solves it. Also solved the issue for my project + a handfull of examples all work as intended.

I also looked into some PRs regarding render mechanism (e.g. #1027), and it seems like you're reworking it, but my change seems to provide good value for what it is, without messing up the API and so on...

I might just be missing some important concept here, so please let me know what you think, thanks!


LeperGnome:
Hey guys! Any feedback? I'll gladly fix remaining tests if you confirm, that this is the right direction.

meowgorithm:
Wow, thanks for this, @LeperGnome. I didn't expect this to work as well as it appears to.

We are indeed working on a new renderer, but this is potentially quite helpful in the meantime as you suggest. That said, it's still a very fundamental change, so we will need to test this very thoroughly in a variety of terminals, environments, and situations before we can merge it (for example, on Windows, behind SSH, and so on). Because of that, I'd like to merge it but can't offer an answer or ETA just yet.

LeperGnome:
@meowgorithm that's great! Let me know if you need any help in testing / fixing golden solution.

meowgorithm:
@LeperGnome yeah, if you could fix the tests that would be awesome!

LeperGnome:
Could you please run the CI pipeline, so I can see if I fixed all the tests

LeperGnome:
All right, checks green, let me know if you need any more help, looking forward to this thing being merged!

meowgorithm:
Thanks for doing that! As I mentioned, this will take some time for us to vet, and I can neither offer and ETA nor promise it will be merged, but so far it seems pretty good.

standard_renderer.go
```diff
@@ -179,64 +184,25 @@ func (r *standardRenderer) flush() {
...
-		// clear the queued message lines
+		// Clear the queued message lines.
 		r.queuedMessageLines = []string{}
 	}
 
-	// Paint new lines
+	// Paint new lines.
 	for i := 0; i < len(newLines); i++ {
-		if _, skip := skipLines[i]; skip {
+		if _, skip := r.ignoreLines[i]; skip {
```
aymanbagabas, line 205: We probably want to keep the skipLine line diff comparison logic as an optimization.

standard_renderer.go
```diff
@@ -268,6 +237,12 @@ func (r *standardRenderer) flush() {
 			}
 		}
 	}
+
+	// Clearing left over content from last render.
+	if r.linesRendered > numLinesThisFlush {
+		buf.WriteString(ansi.EraseDisplayRight)
```
aymanbagabas, line 243: Could you rebase onto `main`?

```suggestion
		buf.WriteString(ansi.EraseDisplayBelow)
```

aymanbagabas (commented):
Hi @LeperGnome, thank you for sending this PR. I think this makes sense to merge, just a few comments below. We also need to ensure that this is VT100 compliant and works with all terminals out there i.e. Linux Console, Urxvt, Xterm, etc.

standard_renderer.go
LeperGnome: I sure can do that, but was there any test, that proved, that this is actually an optimisation? Argument against it can be that emulator rendering the line is faster, then comparing two strings + sometimes rendering anyway.

Decision is yours, but I think that you'll agree, that making code simpler is a good thing.

LeperGnome:
@aymanbagabas 
> ensure that this is VT100 compliant

I'm not a terminal emulator guru by any means, but it seems like control characters, that I've introduced are listed on [vt100.net](https://vt100.net/docs/vt100-ug/chapter3.html). Does it ensure the compliance?

aymanbagabas:
> I'm not a terminal emulator guru by any means, but it seems like control characters, that I've introduced are listed on [vt100.net](https://vt100.net/docs/vt100-ug/chapter3.html). Does it ensure the compliance?

Yes, but I would still test this inside terminals, especially things like the [Linux Console](https://man7.org/linux/man-pages/man4/console_codes.4.html).

LeperGnome:
@aymanbagabas Hi! What should I do about the `skipLine`? As I pointed out in review comment, it's not obvious for me, that this is actually an optimisation (correct me if I'm wrong). If decision was up to me, I would 
- either remove it (as I did in PR) 
- or conduct some testing with different terminal emulators, terminal sizes, rendered content

I'm not really sure, how would I measure results, since I would have to measure not only the speed of bubbletea program frame generation, but also emulator rendering speed with relation to frame content length.

aymanbagabas:
> @aymanbagabas Hi! What should I do about the `skipLine`? As I pointed out in review comment, it's not obvious for me, that this is actually an optimisation (correct me if I'm wrong). If decision was up to me, I would

It would be a huge optimization for remote sessions like SSH. You want to reduce the amount of data sent over the wire as much as possible in such cases. I think comparing strings is not expensive and saves a lot of data from being sent.

```go
if newLines[i] == oldLines[i] {
  buf.WriteString(ansi.CursorDown1) // "\x1b[B"
} else {
  line = line + ansi.EraseLineRight
  // ...
}
```

`ansi.CursorDown1` is only 3 bytes `[\x1b [ B]` compared to the whole line

LeperGnome:
Oh, I never thought of this in context of ssh, I guess you're right!
I'll update my PR.

LeperGnome:
@aymanbagabas Hey, I added skipping logic. 
Also tried to make it work even with queued messages, spent some time debugging why it does work, hours later noticed this:
```golang
case printLineMessage:
    ... 
    r.repaint()
```
So I figured that it's easier to just skip this optimization in this case... Also added a comment, so the next person won't struggle so much =)

mcoops:
I'll throw in, can confirm this works fantastic for our usage, especially in a pty, and it'd be a great stop gap until the rewrite for the renderer is ready.

The last commit especially has removed any random artifacting that we were experiencing - the TUI screen was stable except for some occasional, random refresh for a lack of a better word. 

But it's really nice now!

LeperGnome:
@meowgorithm @aymanbagabas Hey guys, let me know if you need any more help regarding this PR! For now it seems like my part is done.

yorukot:
It works incredibly well and successfully fixed my issue(#1032)! I've tested it on both tmux and zellij, and everything runs perfectly.

LeperGnome:
@yorukot I wrote this fix for my file manager, now it slowly fixes other file managers, such as your `superfile`! Really glad to hear!

meowgorithm:
Hey all! Just a note that this one’s on our radar and we'd like to get it in. Thanks, everyone, for letting us know how this one is working for you.

aymanbagabas (approved):

bashbunni:
Quick note on this one: we're stoked about this and want to test it further with Glow, bubbles, and ensure all examples work properly before merging 🙏

meowgorithm:
Let's also make a point to test this on Windows. In theory it'll be fine, but we should do our diligence there, just in case.

aymanbagabas:
Tested Glow and a some examples on both macOS (Ghostty) and Windows Terminal. All of them run great except for the `package-manager` example. The cursor doesn't go back to the starting position after printing a line.

CC/ @LeperGnome

LeperGnome:
@aymanbagabas took me while, but it seems that somehow `ansi.CursorUp(0)` breaks rendering on single-line frames. This does not make much sense to me, and if you have more insight, please share. Anyhow, I fixed the issue by just not moving up if latest frame was a single line...

LeperGnome:
Ohh... I see... Looking at `ansi.CursorUp(n)` I see this:
```go
func CursorUp(n int) string {
	var s string
	if n > 1 {
		s = strconv.Itoa(n)
	}
	return "\x1b[" + s + "A"
}
```
which basically means that `ansi.CursorUp(0)` is the same as `ansi.CursorUp(1)`. Does not seem too good to me.

LeperGnome:
@meowgorithm checks are failing on main =(

aymanbagabas:
> which basically means that `ansi.CursorUp(0)` (**results in `\x1b[A` (!!!)**) is the same as `ansi.CursorUp(1)` (results in `\x1b[A`, and there's probably no reason to make it a special case, except may be it's shorter). Does not seem too good to me.

Right, the reason behind this is to make it shorter. All of these have the same meaning `\x1b[0A`, `\x1b[1A`, and `\x1b[A` which is to move the cursor up by one.

LeperGnome:
Wait... `\x1b[0A` is that same as `\x1b[A` on the api level? That's wild. I thought it should be a noop or something. Anyway, thank you. You live - you learn xD

meowgorithm:
fwiw @aymanbagabas and I had almost this exact same exchange at some point

LeperGnome:
Hey guys! I see some progress being done on new renderer, which supersedes my changes. Does this mean anything regarding this PR being useful? i.e. are you still planning on merging this?)

meowgorithm:
Hey, @LeperGnome! We're still planning on merging this. We're trying to be a bit careful with it because it is such a critical part of the runtime.

To confirm, did you fix `tea.Println` in Windows https://github.com/charmbracelet/bubbletea/pull/1132#issuecomment-2414681380?

LeperGnome:
@meowgorithm I don't have a Windows machine, but it seems like the issue was not platform-related. And yes, I fixed it in this wonderful big brain commit: https://github.com/charmbracelet/bubbletea/pull/1132/commits/8e3bbbcbdc9db9f6d59e2ca4dcc450f4b0cb1563 =)

standard_renderer.go
```diff
@@ -179,95 +185,69 @@ func (r *standardRenderer) flush() {
...
+		_, _ = buf.WriteString(line)
+
+		if i < len(newLines)-1 {
+			_, _ = buf.WriteString("\r\n")
 		}
 	}
+
+	// Clearing left over content from last render.
+	if r.linesRendered > numLinesThisFlush {
+		buf.WriteString(ansi.EraseDisplayBelow)
```
aymanbagabas, line 248:
```suggestion
		buf.WriteString(ansi.EraseScreenBelow)
```

LeperGnome: bumped `ansi` version in main?

aymanbagabas: Yes, I just released a v1.1.2 to include the `x/ansi` changes

meowgorithm:
Just an update @LeperGnome: things look good to us over here. We’re going to test a little bit more and if things look good we’ll plan to merge this next week.

Thanks for the PR: it’s awesome.

caarlos0 (approved):
changes look good to me.

tested it both standalone and through SSH. Seemed to work great in both cases!

Thanks for the PR, btw ❤️

meowgorithm (approved):

meowgorithm:
It is done! This will be available in the next release. Thanks for this awesome PR, @LeperGnome, and for your patience with it. This is _such_ an improvement for literally everyone across the board.

LeperGnome:
Thank you all! I'm really glad I could make such an improvement for already awesome library! 

- [ ] Build a house
- [ ] Raise a son
- [ ] Plant a tree
- [x] Make an opensource contribution

meowgorithm:
Just a note that this is now available in Bubble Tea v1.2.0. Thanks again for the great contribution!

PR #1139, opened 2024-09-12
(v2) New renderer api

aymanbagabas:
This simplifies the `renderer` interface by adding an `update(Msg) Cmd` function that communicate the program events to the renderer instead of having special cases for each event the renderer needs to know about.

This also adds support for nested `sequenceMsg`s.


screen.go
```diff
@@ -32,43 +32,32 @@ type clearScreenMsg struct{}
...
-
 // EnableMouseCellMotion is a special command that enables mouse click,
 // release, and wheel events. Mouse movement events are also captured if
 // a mouse button is pressed (i.e., drag events).
 //
 // Because commands run asynchronously, this command should not be used in your
 // model's Init function. Use the WithMouseCellMotion ProgramOption instead.
 func EnableMouseCellMotion() Msg {
-	return enableMouseCellMotionMsg{}
+	return sequenceMsg{
```
meowgorithm, line 55: We might want to rename this to something like `ansiSequenceMsg` internally so it doesn't get confused with `Sequence`, the `Cmd`.

aymanbagabas: `sequenceMsg` is the same message type we use with `tea.Sequence`. The idea here is to write the mouse sequences in order.

Issue #1126, opened 2024-08-31
"Alt Gr" + ] generate "esc" instead "]" on french keyboard on Windows.


hypolas:
**Describe the bug**
I use french keyboard. When I try to type this caracter "]", in french keyboard I need to do "Alt Gr" + ].
But in input bubbletea this keys generate "esc" instead "]".
I had tried with full new Windows 11 installation but the problem is same. No problem in Linux.

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS: Windows 11
 - cmd

**To Reproduce**
Run this code:
https://github.com/charmbracelet/bubbletea/blob/main/examples/textinput/main.go
And try to do a "]" in input.

**Source Code**
https://github.com/charmbracelet/bubbletea/blob/main/examples/textinput/main.go

**Screenshots**
![1442547_pastedImage_0](https://github.com/user-attachments/assets/0d627946-93e2-449e-99e5-bf4e167efc15)

hypolas:
Same problem for type "@".

PR #1162, opened 2024-09-19
fix: windows: AltGr maps to LEFT_CTRL+RIGHT_ALT

aymanbagabas:
On Windows, some keyboard layouts have an AltGr button that's similar to the Mac options button and can be used to send characters. This PR respects the AltGr button and treat its characters as bubbletea key runes.

Fixes: https://github.com/charmbracelet/bubbletea/issues/1126

### DIFF
--- a/key_windows.go
+++ b/key_windows.go
@@ -274,7 +274,11 @@ func keyType(e coninput.KeyEventRecord) KeyType {
 	case coninput.VK_DELETE:
 		return KeyDelete
 	default:
-		if e.ControlKeyState&(coninput.LEFT_CTRL_PRESSED|coninput.RIGHT_CTRL_PRESSED) == 0 {
+		switch {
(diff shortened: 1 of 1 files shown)


key_windows.go
```diff
@@ -334,7 +338,7 @@ func keyType(e coninput.KeyEventRecord) KeyType {
 		case '\x1a':
 			return KeyCtrlZ
 		case '\x1b':
-			return KeyCtrlCloseBracket
+			return KeyCtrlOpenBracket // KeyEscape
```
caarlos0, line 341: why the `// KeyEscape`?

caarlos0 (approved):

key_windows.go
aymanbagabas: They're the same value both are `\x1b`

caarlos0 (approved):

meowgorithm (approved):

PR #1168, opened 2024-09-25
feat: use updated table bubble in example

bashbunni:
To try it yourself: `cd examples/table && go run .`

You'll be able to play with the table example using the table bubble rendered with lipgloss.

Dependencies are pinned to the `use-lipgloss-table` branch for both bubbles and lipgloss for validation purposes. The changes to `go.mod` and `go.sum` should be removed before merging.

image
image


bashbunni:
To demo this PR, I have this `go.work` file in my root directory

```
go 1.23.1

use (
	.
	../bubbles
	./examples
)
```

Where `../bubbles` is https://github.com/charmbracelet/bubbles/pull/617
using https://github.com/charmbracelet/lipgloss/pull/373 in the bubbles `go.work` file

Another thing I want to add here is terminal size detection that sets the table height accordingly.

bashbunni:
Demo:

https://github.com/user-attachments/assets/88cab61f-678b-4010-9471-1bd93d816fd3

examples/table/main.go
```diff
@@ -22,6 +22,8 @@ func (m model) Init() tea.Cmd { return nil }
 func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	var cmd tea.Cmd
 	switch msg := msg.(type) {
+	case tea.WindowSizeMsg:
+		m.table.SetHeight(msg.Height - 1)
```
meowgorithm, line 26: The `1` here is a magic number: why is it needed?

bashbunni: hahaha legit no reason at all, just updated it to remove that and it works great

PR #1169, opened 2024-09-27
(v2) keyboard docs

meowgorithm:
This revision expands the keyboard enhancements docs a little to help users find their way around the feature.

One question, @aymanbagabas: is it possible to disable keyboard enhancements on Windows? If not we should adjust the docs here accordingly.

### DIFF
--- a/keyboard.go
+++ b/keyboard.go
@@ -63,8 +63,25 @@ func withKeyDisambiguation(k *keyboardEnhancements) {
 
 type enableKeyboardEnhancementsMsg []KeyboardEnhancement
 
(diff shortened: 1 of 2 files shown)


aymanbagabas:
> One question, @aymanbagabas: is it possible to disable keyboard enhancements on Windows? If not we should adjust the docs here accordingly.

Not really. We could ignore the messages instead, i.e. have a different internal flow than Unix platforms. This happens early in the program when we're initializing the terminal. On Windows, we need to tell the terminal handler that we're using the Windows API vs Virtual Escape Codes.

meowgorithm:
Things are changing in this area and we'll want to revisit it. Closing accordingly.

PR #1187, opened 2024-10-16
docs: add vs code debugging section with headless delve

imyounas:
In the current README.md, the debugging section lacks the instructions on how to setup VS Code with delve in headless mode to debug bubble tea apps. 
I have added step by step instructions on how to do this, believing it will be a great help for new developers  on how to debug CLI apps. 
Kindly review & approve it.
Thanks - Imran

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,33 @@ for backwards compatibility reasons. However, delve recommends using version 2
 for all new development and some clients may no longer work with version 1.
 For more information, see the [Delve documentation](https://github.com/go-delve/delve/tree/master/Documentation/api).
 
+### Debugging with Delve using VS Code
+If you're using VS Code and want to debug your Bubble Tea app with delve, you'll need to run delve in headless mode as mentioned earlier.
+Here are the steps that you can follow to perform headless delve debugging in VS Code:
+- Please install delve following the instructions [here](https://github.com/go-delve/delve/tree/master/Documentation/installation)
+- Please create a new `launch.json` file or add this object under 'configurations' in the existing `launch.json` file.
+ ```json
+            {
+                "name": "Debug Bubble Tea App",
+                "type": "go",
+                "request": "attach",
+                "mode": "remote",
+                "remotePath": "${workspaceFolder}",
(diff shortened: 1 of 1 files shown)


fireflycons:
Nice, however you may want to adjust this to include how to pass command line arguments to the program being debugged. Took me 15-20 min or so to work this out, not having used dlv in this manner before.

I think specifically point out that `args` is invalid in `launch.json` for this remote mode, and that the args need to be passed to dlv directly - that everything after `--` is passed to the program being debugged, e.g.

```text
dlv debug --headless --api-version=2 --listen=127.0.0.1:43000 . -- arg1 --arg2 [ etc ]
```

aymanbagabas (commented):
Perhaps in `DEBUG.md` Wdyt @charmbracelet/everyone?

raphamorim:
> Perhaps in `DEBUG.md` Wdyt @charmbracelet/everyone?

I think is better in a DEBUG file, so we can also add more information not only for VS Code users but other editors/stacks

raphamorim (approved):

bashbunni:
Personally, I think it makes sense to link to their own editor's documentation instead of re-writing it in our own words then having to maintain that copy. We can link to the VS Code documentation on how to connect to a headless delve server https://github.com/golang/vscode-go/blob/master/docs/debugging.md#remote-debugging

I don't have any strong opinions about keeping it in the README vs DEBUG.md file

README.md
```diff
@@ -283,6 +283,33 @@ for backwards compatibility reasons. However, delve recommends using version 2
 for all new development and some clients may no longer work with version 1.
 For more information, see the [Delve documentation](https://github.com/go-delve/delve/tree/master/Documentation/api).
 
+### Debugging with Delve using VS Code
```
bashbunni, line 286:
```suggestion
### Connecting to a headless session

You can connect your editor's debugger to a headless delve session. This is known as remote debugging.

- [VSCode - Remote Debugging](https://github.com/golang/vscode-go/blob/master/docs/debugging.md#remote-debugging)
```

I think we could just do this and remove the other instructions below

PR #1194, opened 2024-10-22
(v2) fix: make grapheme clustering opt-in

aymanbagabas:
This commit changes the behavior of grapheme clustering to be opt-in instead of opt-out. This is because some terminals like Apple Terminal don't support DECRQM and DECRPM, which are required for querying and setting the grapheme clustering state.

Apple Terminal, instead, replies with a `p` character when querying the grapheme clustering state, which is not a valid response and breaks the program output.

image

Notice the `p` in the first line


bashbunni (approved):

meowgorithm (approved):
Well, it's cool that the feature exists in an opt-in fashion, at least. LGTM.

PR #1195, opened 2024-10-22
(v2) feat: query background color by default

aymanbagabas:
This commit makes the program query the terminal for its background color by default. This is useful if you want to enable automatic background color detection for your program.

It also adds a new option `WithoutBackgroundColor` to disable this behavior.


meowgorithm:
For the alpha, why don't we keep it opt in? If we find the friction is too high we can merge this. To your point, it would be good to avoid the heavy query when we can.

bashbunni:
Aren't users used to this behaviour by default though? I do agree that less is more overall though. Just might feel weird at first if they're used to a certain behaviour

aymanbagabas:
I agree with @bashbunni tbh. Since we're moving to pure lipgloss and removing the lipgloss "hack" in bubbletea@v2 that queries the terminal on start, we need to send the terminal bg color on start similar to `tea.WindowSizeMsg`.

bashbunni (approved):

caarlos0 (approved):

meowgorithm:
Because checking for the background color is so much more involved now (you need to listen for `tea.BackgroundColorMsg` in the same way you listen for `tea.WindowSizeMsg`) I don't think it's to much to ask to have users explicitly fire off the request. That said, if we find during the alpha/beta that people are getting confused we can enable it by default. It's easier to add versus take it away.

bashbunni:
What's the benefit of having the user fire off that request? Aren't they querying the terminal for its environment info for the WindowSizeMsg anyway?

meowgorithm:
Superseded elsewhere: closing.

PR #1211, opened 2024-10-28
(v2) feat: add WithRequestBackground option to query terminal background c…

aymanbagabas:
…olor

This commit adds a new option to the program, `WithRequestBackground`, which will send a request to the terminal to query the background color of the terminal.


caarlos0 (approved):

aymanbagabas:
I don't think this option makes sense, as it would create two different workflows to request the background color, one through calling `tea.RequestBackgroundColor` in `Init` and using this option in when creating the program.

PR #1215, opened 2024-10-29
feat: Wrap error returned from run with context error

### LINKED ISSUE #1212: context error is swallowed up, preventing errors.Is
**Describe the bug**

**Setup**
 - OS Any
 - Shell Any
 - Terminal Emulator Any
 - Terminal Multiplexer Any

**To Reproduce**
See source code

**Source Code**
```go
_, err := p.Run()
  // Handle timeout
}
```

**Expected behavior**

**Additional context**

goodevilgenius:
Fixes #1212 

This change allows the error returned by `p.Run` to be checked for a context error.

For example, given this code:

```go
_, err := p.Run()
  // Handle timeout
}
```

This change now allows this to work.

This opens up more possibilities, allowing a program to cancel a context, and look for the reason for the cancellation.

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -595,7 +595,7 @@ func (p *Program) Run() (Model, error) {
 	model, err := p.eventLoop(model, cmds)
 	killed := p.ctx.Err() != nil
 	if killed {
-		err = fmt.Errorf("%w: %w", ErrProgramKilled, context.Cause(p.ctx))
+		err = fmt.Errorf("%w: %w", ErrProgramKilled, p.ctx.Err())
 	} else {
 		// Ensure we rendered the final state of the model.
 		p.renderer.write(model.View())
--- a/tea.go
+++ b/tea.go
@@ -595,7 +595,7 @@ func (p *Program) Run() (Model, error) {
 	model, err := p.eventLoop(model, cmds)
 	killed := p.ctx.Err() != nil
 	if killed {
-		err = fmt.Errorf("%w: %s", ErrProgramKilled, p.ctx.Err())
+		err = fmt.Errorf("%w: %w", ErrProgramKilled, p.ctx.Err())
 	} else {
 		// Ensure we rendered the final state of the model.
 		p.renderer.write(model.View())
--- a/tea.go
+++ b/tea.go
@@ -595,6 +595,7 @@ func (p *Program) Run() (Model, error) {
 	model, err := p.eventLoop(model, cmds)
 	killed := p.ctx.Err() != nil
 	if killed {
+		// TODO: change p.ctx.Err() to context.Cause(p.ctx) on upgrade to go 1.20 to take into account context.WithCancelCause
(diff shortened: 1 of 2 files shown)


goodevilgenius:
Ooh, I see `go.mod` is pinned to version 1.18. That makes `context.Cause` unavailable. That's unfortunate. I'll revert that commit, so it's compatible with 1.18.

meowgorithm:
Wow, that was fast! If functionality is the same with the new version, want to also put a little comment in there about how to use the hot, new `context.Cause` when we're ready to bump the lib up to 1.20?

goodevilgenius:
@meowgorithm Sure. Not a problem. What's the standard way to do that on this project? A `// TODO` comment above the code, or some other method?

meowgorithm:
`// TODO:` is just fine!

tea.go
```diff
@@ -595,7 +595,7 @@ func (p *Program) Run() (Model, error) {
 	model, err := p.eventLoop(model, cmds)
 	killed := p.ctx.Err() != nil
 	if killed {
-		err = fmt.Errorf("%w: %s", ErrProgramKilled, p.ctx.Err())
+		err = fmt.Errorf("%w: %w", ErrProgramKilled, p.ctx.Err())
```
meowgorithm, line 598: Wow, TIL you cannot double-wrap

goodevilgenius: Yeah, it does something similar to `errors.Join` when you double wrap errors. This *does* break `errors.Unwrap`, since that function only works on singly-wrapped errors, but `errors.Is` and `errors.As` continues to work as expected.

goodevilgenius:
Darnit. Looks like this is not possible with this version of go. Pipeline is still failing with:
> fmt.Errorf call has more than one error-wrapping directive %w

It works fine on my machine, running 1.23.

I checked back with the old docs. `fmt.Errorf` says this for 1.18:

> It is invalid to include more than one %w verb

But, for 1.20 and after, it says:

> If there is more than one %w verb, the returned error will implement an Unwrap method returning a []error containing all the %w operands in the order they appear in the arguments.

Looks like this is a non-starter as long as bubbletea is based on 1.18.

I see that the 2.0 alpha branch is still on 1.18, and not a newer version. Any thoughts on when this upgrade might happen?

meowgorithm:
Maybe it's time. I know a lot of projects typically support the last two versions of Go, and 1.18 came out about two and a half years ago. Want to reopen this?

goodevilgenius:
@meowgorithm Should I reopen it targeting the 2.0 branch, instead of main?
I imagine bumping the version of go would probably constitute a breaking change, no?

meowgorithm:
v2 would be ideal, however are you okay working against that version, seeing that it will take some time for us to reach a stable release?

meowgorithm:
Sounds good. This will be a good one to get in, so thank you.

goodevilgenius:
I'll rebase against the v2 branch, and open a new one. I can't find a way to change the base on GitHub anyway.

PR #1233, opened 2024-11-11
fix: incorrect line skipping in renderer flush

### LINKED ISSUE #1232: Incorrect line skipping in renderer flush
**Describe the bug**

In the renderer `flush`, lines are skipped if they match the previous line. However, the comparison between the current and previous lines is not done properly.

```go
newLines := strings.Split(r.buf.String(), "\n")
oldLines := strings.Split(r.lastRender, "\n")
```

After that, the `newLines` are trimmed based on the terminal height to fit the screen:

```go
if r.height > 0 && len(newLines) > r.height {
    newLines = newLines[len(newLines)-r.height:]
}
```

This trimming causes the index to be lost. The first item in `oldLines` may not correspond to the first item in `newLines` if the number of lines in the current buffer exceeds the terminal height.

The issue occurs in the following comparison:

```go
canSkip := !flushQueuedMessages && 
    len(oldLines) > i && oldLines[i] == newLines[i]
```

In some cases, this leads to skipping line updates even when two lines are not the same (and vice versa). This happens because `oldLines` always starts from the top of the previous buffer, while `newLines` may have been truncated.

The solution is to compare the most recently rendered lines with the new lines, instead of comparing the lines from the beginning of the previous buffer.

**Setup**
 - OS: Debian
 - Shell: zsh
 - Terminal Emulator: kitty
 - Terminal Multiplexer: tmux

**To Reproduce**

It can be reproduced with the following steps:

1. Press `j` to increase the number of lines until it exceeds the terminal height.
2. Press `k` to decrease the number of lines until it is smaller than the terminal height.
3. You will observe that the lines are misaligned, with each line appearing one row higher than its correct position on the screen, and the first line will be missing.

```go
package main

import (
	"fmt"
	"os"
	"strings"

)

type model struct {
	n int
}

	return nil
}

	switch msg := msg.(type) {
	case tea.KeyMsg:
		switch msg.String() {
		case "q", "ctrl+c", "esc":
			return m, tea.Quit
		case "j":
			m.n += 1
		case "k":
			m.n -= 1
		}
	}
	return m, nil
}

	b := strings.Builder{}
	for i := range m.n {
		b.WriteString(fmt.Sprintf("%d. line\n", i))
	}
	return fmt.Sprintf("FirstLine\n%sLastLine", b.String())
}

func main() {
	if _, err := tea.NewProgram(model{n: 9}).Run(); err != nil {
		fmt.Println("Error running program:", err)
		os.Exit(1)
	}
}

```

**Expected behavior**

Lines should be skipped correctly only when the current line is identical to the previously rendered line

**Screenshots**
Screenshot from 2024-11-11 23-03-03
Adding lines...
Screenshot from 2024-11-11 23-03-20
Deleting lines...
Screenshot from 2024-11-11 23-03-34

semihbkgr:
fixes #1232

To fix the bug described in the issue, we can simply store the currently rendered lines and use them for comparison in the upcoming flush operation.

### DIFF
--- a/standard_renderer.go
+++ b/standard_renderer.go
@@ -184,7 +184,6 @@ func (r *standardRenderer) flush() {
 		newLines = newLines[len(newLines)-r.height:]
 	}
 
-	numLinesThisFlush := len(newLines)
 	flushQueuedMessages := len(r.queuedMessageLines) > 0 && !r.altScreenActive
 
 	if flushQueuedMessages {
@@ -257,11 +256,11 @@ func (r *standardRenderer) flush() {
 	}
 
 	// Clearing left over content from last render.
-	if r.linesRendered > numLinesThisFlush {
+	if r.linesRendered > len(newLines) {
 		buf.WriteString(ansi.EraseScreenBelow)
 	}
 
-	r.linesRendered = numLinesThisFlush
+	r.linesRendered = len(newLines)
 
 	// Make sure the cursor is at the start of the last line to keep rendering
 	// behavior consistent.
--- a/standard_renderer.go
+++ b/standard_renderer.go
@@ -34,6 +34,7 @@ type standardRenderer struct {
 	ticker             *time.Ticker
 	done               chan struct{}
 	lastRender         string
+	lastRenderedLines  []string
 	linesRendered      int
 	useANSICompressor  bool
 	once               sync.Once
@@ -174,7 +175,6 @@ func (r *standardRenderer) flush() {
 	}
 
 	newLines := strings.Split(r.buf.String(), "\n")
-	oldLines := strings.Split(r.lastRender, "\n")
 
 	// If we know the output's height, we can use it to determine how many
 	// lines we can render. We drop lines from the top of the render buffer if
@@ -210,7 +210,7 @@ func (r *standardRenderer) flush() {
 	// Paint new lines.
 	for i := 0; i < len(newLines); i++ {
 		canSkip := !flushQueuedMessages && // Queuing messages triggers repaint -> we don't have access to previous frame content.
-			len(oldLines) > i && oldLines[i] == newLines[i] // Previously rendered line is the same.
+			len(r.lastRenderedLines) > i && r.lastRenderedLines[i] == newLines[i] // Previously rendered line is the same.
 
 		if _, ignore := r.ignoreLines[i]; ignore || canSkip {
 			// Unless this is the last line, move the cursor down.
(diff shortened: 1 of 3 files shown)


semihbkgr:
It seems that the bug was introduced in #1132 after this change:
https://github.com/charmbracelet/bubbletea/pull/1132/files#diff-eec1160b5c25625c859865f3ea86e45e76521a2dd877c78a508183b968646f7dL194

meowgorithm:
Totally, this is a regression. Appreciate all the work and insight on this one, @semihbkgr; we’re checking this out now.

aymanbagabas (commented):
I tested it with the provided example and everything seems to work as expected.

@semihbkgr Could you please add comments in the code explaining why you're doing this? I know you explained this in the issue, but it would be more beneficial to have some explanation in the code 🙂

LeperGnome:
I could reproduce the bug, yea

LeperGnome:
@meowgorithm I found out that this bug was present in the original renderer before 1.2.0, but in a slightly different form... I'm tinkering around, still can't really wrap my head around it. I managed to fix it myself by:
```go
	r.lastRender = strings.Join(newLines, "\n") // instead of r.buf.String()
```
But I didn't test current patch yet

LeperGnome:
Yea, as @semihbkgr pointed out in https://github.com/charmbracelet/bubbletea/issues/1232 it seems like the issue lies in the fact, that `lastRender` contains **all** the lines (even not rendered) from the previous frame, and `newLines` are being trimmed to fit the window height. So at the end, we don't match lines correctly to see if they're the same. ouch

LeperGnome:
Oh, wait.. @aymanbagabas did you reproduce the bug or not? I might've misunderstood your comment

aymanbagabas:
@LeperGnome Yes, I was able to reproduce the bug. This patch fixes the issue

standard_renderer.go
```diff
@@ -34,6 +34,7 @@ type standardRenderer struct {
 	ticker             *time.Ticker
 	done               chan struct{}
 	lastRender         string
+	lastRenderedLines  []string
 	linesRendered      int
```
semihbkgr, line 38: We can possibly remove the `linesRendered` field and use `len(lastRenderedLines)` instead. What do you think?

aymanbagabas: It will break `tea.Println`. We reset the `lastRenderedLines` when we call `repaint`, which is the case when using `tea.Println`. We need to keep track of the number of lines last rendered for `tea.Println` to work.

LeperGnome:
As a side note - maybe it's a good thing to introduce tests for such corner-cases? I'll definitely help when working on new renderer.

PR #1234, opened 2024-11-12
docs: clarify imports in the readme

caarlos0:


bashbunni:
I think this is fine to include. I haven't seen any questions about this before and it may cause more confusion for new users than it does to clarify uncertainty 🤔

PR #1240, opened 2024-11-16
Update main.go - implement missing getPackages

7flash:
before it shows

```
examples/package-manager/main.go:41:13: undefined: getPackages
```

now it works

Screenshot 2024-11-16 at 10 43 30

### DIFF
--- a/examples/package-manager/main.go
+++ b/examples/package-manager/main.go
@@ -132,6 +132,20 @@ func max(a, b int) int {
 	return b
 }
 
+func getPackages() []string {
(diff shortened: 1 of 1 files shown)


meowgorithm:
Hi! Thank you for the thoughtful PR. `getPackages` is actually implemented, it's just in [another file in the directory](https://github.com/charmbracelet/bubbletea/blob/76b0f818b3174a3e9b77f93c19820d07855f13f4/examples/package-manager/packages.go). To run the example you'll want to `cd` into the `package-manager` directory and `go run .`.

```bash
cd ./examples/package-manager
go run .
```

If we've missed something here let us know and we'll reopen. Thanks!

Issue #1242, opened 2024-11-17
Improve the realtime example by passing some data over the channel


mdepot:
This is to suggest a minor improvement to the realtime example.  Currently the /examples/README.md file says it "demonstrates the use of go channels to perform realtime communication with a Bubble Tea application."

Right now a channel is in use, however no data is being passed over it, so it's not really demonstrating how to update the model with something that was passed over a channel.

Instead of just counting the number of times that no data was sent, this example could be improved if it actually passed some sort of data in the structure, and then Update() could update the model with whatever was received through the channel, and then the View() could actually show that data as successfully arrived.

meowgorithm:
That sounds good to me. Feel free to open a PR to add that functionality.

FFCoder:
@meowgorithm  Hey! I created a PR for this. Please let me know your thoughts! 🙂

PR #1243, opened 2024-11-19
Updated Realtime Example

FFCoder:
Per Issue #1242 updated the real time example to grab some random chat messages and send it down the channel. This will be used to demonstrate how a production application could send messages over a channel. 

Largely, I tried to limit changes as much as possible other than what is needed. I'm still new-ish to Golang so I welcome criticism or if we want to go another way with this I'm happy to change!


mdepot:
Thanks, this is already helpful.  Prior to this there were a number of places where the channel was just defined as struct{}.  Now it's obvious that responsMsg is the type that should be on the channel, and that is key to understanding this example.

mdepot:
In retrospect, since responseMsg was already defined then just substituting `chan responseMsg` for all the `chan struct{}` and `sub <- responseMsg{}` for `sub <- struct{}{}` might have made it clear enough, even without passing any data inside the responseMsg.

mdepot (changes requested):
It looks like this would still work ok if ChatMessage was private (lower case).  If not, please help me understand why it might be good to leave it exported.

Also, printing Last Message could use a leading space for alignment.

FFCoder:
@meowgorithm  or @aymanbagabas  would either of you mind reviewing this?

meowgorithm:
Hey @FFCoder. Apologies for the delay on this one: we've been very backlogged. This is a helpful PR and wed be happy to merge it with some changes (notably, I'd use the `tea.Println()` command to print messages).

If you're willing to reopen it we'd be happy to make those changes and merge it.

Issue #1248, opened 2024-11-22
(v1) tea.EnterAltScreen to tea.ExitAltScreen will flash or fail to exit and only render parts


unitsvc:
**Describe the bug**

(v1.2.3) tea.EnterAltScreen to tea.ExitAltScreen will flash or fail to exit and only render parts

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS [Ubuntu, macOS]
 - Shell [zsh, bash]

**To Reproduce**
1. With the table component, the table is wider than the screen

2. Clean screen 

3. EnterAltScreen then ExitAltScreen

meowgorithm:
Thank you very much for the report, @unitsvc. Any chance you can share some source code to help us reproduce the issue?

unitsvc:
@meowgorithm 

```go
package main

import (
	"github.com/charmbracelet/bubbles/table"
	tea "github.com/charmbracelet/bubbletea"
	"github.com/charmbracelet/lipgloss"
)

var baseStyle = lipgloss.NewStyle().
	BorderStyle(lipgloss.NormalBorder()).
	BorderForeground(lipgloss.Color("240"))

func NewTable() table.Model {
	rows := []table.Row{
		{"1", "issue", "v1.2.3", "24/11/22", "EnterAltScreen"},
	}
	columns := []table.Column{
		{Title: "ID", Width: 4},
		{Title: "NAME", Width: 28},
		{Title: "VERSION", Width: 18},
		{Title: "DATE", Width: 15},
		{Title: "REMARK", Width: 4},
	}
	t := table.New(
		table.WithColumns(columns),
		table.WithRows(rows),
		table.WithFocused(true),
		table.WithHeight(7),
	)
	s := table.DefaultStyles()
	s.Header = s.Header.
		BorderStyle(lipgloss.NormalBorder()).
		BorderForeground(lipgloss.Color("240")).
		BorderBottom(true).
		Bold(false)
	s.Selected = s.Selected.
		Foreground(lipgloss.Color("229")).
		Background(lipgloss.Color("57")).
		Bold(false)
	t.SetStyles(s)
	return t
}

type Model struct {
	currentModel   table.Model
	enterAltScreen bool
}

func (m Model) Init() tea.Cmd {
	if m.enterAltScreen {
		return tea.EnterAltScreen
	}
	return nil
}

func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	var cmds []tea.Cmd
[...]

PR #1241, opened 2024-11-16
fix: cursor position adjustment after exiting alt screen

semihbkgr:
When exiting alternate screen mode, the cursor was not properly adjusted, which caused terminal content to be either overwritten (if the current line count was greater than before entering alt screen) or leftover lines to remain at the top (if the current line count was smaller). To fix this, we save the number of lines rendered before entering alternate screen mode. When exiting, we compare the current line count with the saved value and adjust the cursor position accordingly, ensuring correct positioning and preventing both overwriting and leftover lines.

fixes #1013

Issue #1013, opened 2024-05-10
Exiting fullscreen glitch when also removing a line in output


applejag:
**Describe the bug**
When conditionally rendering a line if you're full screen, then when you exit full screen it clears too many lines and removes terminal output that isn't part of 

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS: Linux, NixOS
 - Shell: tested and getting the same glitch in bash, zsh, fish, and pwsh
 - Terminal Emulator: only tested alacritty, kitty, and iterm
 - Terminal Multiplexer: tested both with and without tmux

**To Reproduce**
1. Have some history in your terminal
2. `go run .`
3. click <kbd>f</kbd> to enter fullscreen and then again to exit fullscreen (doesn't have to be double clicked quickly. The glitch is not time sensitive at all actually)
4. repeat step 3 and see how it clears 1 line at a time from your terminal history

**Source Code**
```go
package main

import (
	"fmt"
	"os"

	tea "github.com/charmbracelet/bubbletea"
)

func main() {
	p := tea.NewProgram(model{})
	if _, err := p.Run(); err != nil {
		fmt.Printf("Alas, there's been an error: %v", err)
		os.Exit(1)
	}
}

type model struct {
	isFullscreen bool
}

func (m model) Init() tea.Cmd {
	return nil
}

func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	switch msg := msg.(type) {
	case tea.KeyMsg:
		switch msg.String() {

		// These keys should exit the program.
		case "ctrl+c", "q":
			return m, tea.Quit

		// Toggle fullscreen
		case "f":
			m.isFullscreen = !m.isFullscreen
			if m.isFullscreen {
				return m, tea.EnterAltScreen
			} else {
				return m, tea.ExitAltScreen
			}
		}
	}
	return m, nil
}

func (m model) View() string {
	s := "Hello\n"

	if m.isFullscreen {
		s += "yay we're fullscreen\n"
	}

	s += "\nPress q to quit.\n"

	return s
}
```

**Expected behavior**
When exiting full screen mode, tea renders the view first before trying to figure out how many lines to clear.

**Screenshots**
When first just running the program:

![image](https://github.com/charmbracelet/bubbletea/assets/2477952/ecbbdd46-4f52-41d9-bd67-e5a097a6315a)

After clicking <kbd>f</kbd> to enter fullscreen:

[...]

PR #1251, opened 2024-11-22
fix: renderer: keep a separate count of lines rendered in the alt screen 

aymanbagabas:
We need to keep a separate count of lines rendered in the alt screen and
inline mode. This is because when we enter alt screen mode, the cursor
position is saved, and the alt screen is cleared. Now when we exit the
alt screen mode, the cursor position is restored so we don't need to
move the cursor to the beginning of the section that we rendered.

To fix this, we keep a separate count of lines rendered in the alt screen
mode and inline mode.

Related: https://github.com/charmbracelet/bubbletea/issues/1013
Fixes: https://github.com/charmbracelet/bubbletea/pull/1241
Fixes: https://github.com/charmbracelet/bubbletea/issues/1248

CC/ @unitsvc @semihbkgr

### DIFF
--- a/screen_test.go
+++ b/screen_test.go
@@ -24,7 +24,7 @@ func TestClearMsg(t *testing.T) {
 		{
 			name:     "altscreen_autoexit",
 			cmds:     []Cmd{EnterAltScreen},
-			expected: "\x1b[?25l\x1b[?2004h\x1b[?1049h\x1b[2J\x1b[H\x1b[?25l\rsuccess\x1b[0K\r\n\x1b[0K\x1b[2;0H\x1b[2K\r\x1b[?2004l\x1b[?25h\x1b[?1002l\x1b[?1003l\x1b[?1006l\x1b[?1049l\x1b[?25h",
+			expected: "\x1b[?25l\x1b[?2004h\x1b[?1049h\x1b[2J\x1b[H\x1b[?25l\x1b[H\rsuccess\x1b[0K\r\n\x1b[0K\x1b[2;0H\x1b[2K\r\x1b[?2004l\x1b[?25h\x1b[?1002l\x1b[?1003l\x1b[?1006l\x1b[?1049l\x1b[?25h",
 		},
 		{
 			name:     "mouse_cellmotion",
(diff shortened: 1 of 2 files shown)


unitsvc:
Thank you for the fix, it works perfectly.

unitsvc:
@meowgorithm CC

semihbkgr:
@aymanbagabas The bug described in #1248 is not a regression in `v1.2.3`. It is a type of bug I intended to fix in my PR, but the scenario explained in the issue (where the program starts in `altscreen` mode) is not covered by my fix.

semihbkgr:
I did some debugging to understand why this case is not covered. It turns out that when the renderer is initialized, `renderedLines` is set to 0 (the default value for an integer) before calling flush. In this scenario, the program enters alt screen mode before `flush` is called for the first time, so `linesRenderedBeforeAltScreen` remains 0. When this value is 0, we avoid moving the cursor in the `exitAltScreen` function.

```go
if r.linesRendered < r.linesRenderedBeforeAltScreen {
...
} else if r.linesRendered > r.linesRenderedBeforeAltScreen && r.linesRenderedBeforeAltScreen > 0 {
...
}
```

So, the default value of `renderedLines` should be set to 1 before calling flush. This is because, when the program starts, the cursor position will be on the first line.

semihbkgr (approved):
LGTM 👍 It’s better and cleaner to use a separate line count for inline and alt screen modes, as this eliminates the need to move the cursor up and down in the `exitAltScreen` function.

semihbkgr:
When testing this PR, I realized that aligning the cursor position by moving it up and down in the `exitAltScreen` function was not a good idea, as it could cause undesired behavior in some cases. For example:
When the program is terminated while in alt screen mode, it calls the `restoreTerminalState` function, which subsequently calls `exitAltScreen`. During this process, the cursor could potentially move down unnecessarily, adding extra unwanted spaces that won’t be used by the next flush, before the program terminates.
The second problem was that moving the cursor down did not work when it reached the bottom of the terminal. The `flush` function can handle this more properly by writing `\r\n`.
These are the issues that will be resolved after this PR.

unitsvc:
@aymanbagabas Can this pr be merged and a new version released

aymanbagabas:
@unitsvc This is now part of v1.2.4

PR #1252, opened 2024-11-23
Minimise quit state checking in help example

ChausseBenjamin:
Currently, every time the view is rendered, the app checks for the
quitting boolean which seems unnecessary. This approach simply returns
a unique view for quitting which removes the need for a `quitting` boolean
variable an the `if quitting {...}` check currently run  everytime
`model.View()` gets called.

### DIFF
--- a/examples/help/main.go
+++ b/examples/help/main.go
@@ -126,10 +126,8 @@ func (m model) View() string {
 
 type quitModel struct{}
 
-func (m quitModel) Init() tea.Cmd { return nil }
-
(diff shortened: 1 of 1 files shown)


caarlos0 (approved):

bashbunni:
Hey @ChausseBenjamin thank you so much for your contribution! We really appreciate you looking at how the examples could be improved. That said, I think this change introduces more complexity than the current implementation. Given that, we are going to close this one. Please don't be shy to contribute again in the future, it's always great to see things from a fresh perspective :)

PR #1254, opened 2024-11-25
fix: getting last rendered line counts including alt screen lines

semihbkgr:
I think it was overlooked in #1251

@aymanbagabas


aymanbagabas (approved):
Thank you @semihbkgr!

PR #1272, opened 2024-12-16
fix: Insert comma to avoid possible missundertood.

locnnil:
When read without the comma, it gives me the impression that there are two tutorials covering the topic of commands.  

(I instantly go back, thinking, "Wait, what did I miss? Where is the first one about commands?")  

However, the intention is to clarify that this is the second tutorial in the series of Bubble Tea tutorials.


meowgorithm:
Wow, great call. Thank you!

PR #1285, opened 2025-01-09
feat(examples): add splash example

aymanbagabas:
This is ported from Textualize splash.py example.

image


bashbunni:
Looks great! Just tried in Ghostty and rio :)

PR #1292, opened 2025-01-21
test: add a test for canceling the context

caarlos0:
this adds test for canceling the context within the bubble


aymanbagabas (approved):

desertwitch:
This PR should be obsolete and resolved with the proposed changes in #1372.

The behavior documented in this PR:
```
// Note that canceling the context from within bubbletea (e.g. in an [Cmd]) is
// not supported.
```
Actually is a symptom of a deadlock issue described in #1370 and resolved through #1372 (when merged).

PR #1293, opened 2025-01-22
(v2) refactor: remove moving the cursor logic

aymanbagabas:
This removes moving the cursor and setting the cursor style from the library until we have a better way to handle it.


meowgorithm:
Does it also make sense to include cursor show/hide commands here?

Issue #1136, opened 2024-09-11
(v2) Proposal: program execution flow


aymanbagabas:
Currently, Bubble Tea programs has the following methods to start and run a program.
- `Start() error` starts the program and waits without returning a final model **deprecated**
- `Run() (Model, error)` starts the program and waits
- `Wait()` waits for the program to finish (the aforementioned ones already wait for the program to exit, this is redundant)
- `Kill()` kills the program (currently it only cancels the context) Related issues: https://github.com/charmbracelet/bubbletea/issues/234 https://github.com/charmbracelet/bubbletea/issues/1127

To keep the API similar to other Go packages, we could change the API to be as follows:
- `Start() error` starts the program and event loop
- `Wait() error` waits for the event loop to finish and exit
- `Run() error` becomes a helper that starts and waits for the program
- `Kill() error` kills the program and forces the event loop to stop

With this, we would expose a `program.Model` that has the final model after the program exits. During the lifetime of the program, `program.Model` would be `nil`. This is similar to the `exec.Cmd` API in the standard library.

tauraamui:
Presumably this would result in needing to provide a V2 version of the library, how would you most prefer to do this? Via a V2 branch or a v2 submodule?

caarlos0:
@aymanbagabas the only problem I see with this is that people might still be using the currently deprecated `Start()`, which waits for the program to quit.

If we do this new proposed API, `Start` becomes "undeprecated", with different behavior. 
Not sure how fine this is, but would def have to be very specific about this.
Another idea would be to make a "tealint" of sort, which check for this kind of stuff (and maybe more?) - seems like a rather big project though.

caarlos0:
> Presumably this would result in needing to provide a V2 version of the library, how would you most prefer to do this? Via a V2 branch or a v2 submodule?

Definitely for v2, haven't decided on specifics yet, as it also depends on whether we'll want to continue maintaining v1, e.g., with GoReleaser I simply said "main is now v2", changed the module paths, and that was it.

dolmen:
To go further, here is another design where the Program is given directly the Init function that instantiate the model which then has only two functions: `Update` and `View`.

```go
type Model[State any] struct {
	// Update is called when a message is received. Use it to inspect messages
	// and, in response, update the model and/or send a command.
	Update func(State) (State, Cmd)

	// View renders the program's UI, which is just a [fmt.Stringer]. The view
	// is rendered after every Update.
[...]

PR #1298, opened 2025-01-27
(v2) Generic model

aymanbagabas:


aymanbagabas:
Here's an example of rewriting the `cursor-style` example using this API

```go
package main

import (
	"fmt"
	"os"

	tea "github.com/charmbracelet/bubbletea/v2"
)

type model struct {
	shape tea.CursorShape
	blink bool
}

func (m model) describeCursor() string {
	var adj, noun string

	if m.blink {
		adj = "blinking"
	} else {
		adj = "steady"
	}

	switch m.shape {
	case tea.CursorBlock:
		noun = "block"
	case tea.CursorUnderline:
		noun = "underline"
	case tea.CursorBar:
		noun = "bar"
	}

	return fmt.Sprintf("%s %s", adj, noun)
}

func initialModel() (model, tea.Cmd) {
	m := model{blink: true}
	return m, nil
}

func updateModel(m model, msg tea.Msg) (model, tea.Cmd) {
	switch msg := msg.(type) {
	case tea.KeyPressMsg:
		switch msg.String() {
		case "ctrl+q", "q":
			return m, tea.Quit
		case "h", "left":
			if m.shape == tea.CursorBlock && m.blink {
				break
			}
			if m.blink {
				m.shape--
			}
			m.blink = !m.blink
		case "l", "right":
			if m.shape == tea.CursorBar && !m.blink {
				break
			}
			if !m.blink {
				m.shape++
			}
			m.blink = !m.blink
		}
	}
	return m, nil
}

func viewModel(m model) fmt.Stringer {
	f := tea.NewFrame("Press left/right to change the cursor style, q or ctrl+c to quit." +
		"\n\n" +
		"  <- This is the cursor (a " + m.describeCursor() + ")")
	f.Cursor = tea.NewCursor(0, 2)
	f.Cursor.Shape = m.shape
	f.Cursor.Blink = m.blink
	return f
}

func main() {
	p := tea.Program[model]{
		Init:   initialModel,
		Update: updateModel,
		View:   viewModel,
	}
	if err := p.Run(); err != nil {
		fmt.Fprintf(os.Stderr, "Error: %v", err)
		os.Exit(1)
	}
}
```

tea.go
```diff
@@ -281,66 +293,83 @@ func Interrupt() Msg {
...
-		initialModel: model,
-		msgs:         make(chan Msg),
-		rendererDone: make(chan struct{}),
-		keyboardc:    make(chan struct{}),
-		modes:        ansi.Modes{},
-	}
+func NewProgram[T any](model Model[T]) *Program[T] {
+	p := new(Program[T])
+	p.Init = model.Init
+	p.Update = func(t T, msg Msg) (T, Cmd) { return any(t).(Model[T]).Update(msg) }
```
dolmen, line 299: This line would me much simpler if `Model[T]` was a struct instead of an interface. See https://github.com/charmbracelet/bubbletea/issues/1136#issuecomment-2754104666

dolmen (changes requested):
The NewProgram implementation wraps the `Update` and `View` functions (which are called frequently in the render loop) in a costly way. See my proposal for an alternate representation of `Model[T]` in https://github.com/charmbracelet/bubbletea/issues/1136#issuecomment-2754104666.

meowgorithm:
I still absolutely love this design. For now, however, in the spirit of keeping things tidy, I'm closing this PR.

PR #1304, opened 2025-02-03
docs: viewport horizontal scroll, gutter, highlight examples

caarlos0:
example usage of https://github.com/charmbracelet/bubbles/pull/697


examples/chat/main.go
```diff
@@ -55,6 +55,8 @@ func initialModel() model {
 	vp := viewport.New(30, 5)
 	vp.SetContent(`Welcome to the chat room!
 Type a message and press Enter to send.`)
+	vp.KeyMap.Left.SetEnabled(false)
+	vp.KeyMap.Right.SetEnabled(false)
```
caarlos0, line 59: not sure if this should be disabled by default?

aymanbagabas: hmm good question. I'd look at existing tools like `less` and `more`

meowgorithm:
I'm playing with this a bit and it looks like not all matches are being highlighted and the highlighting jumps a bit (highlights here are in green, with the selected one a darker green. Unless I'm missing something?

https://github.com/user-attachments/assets/29359c69-0c64-4324-9905-ee34b2925969

caarlos0:
@meowgorithm looks like you only set the selected highlight style, there's a highlight style as well

meowgorithm:
Btw, this is a solid API. Nice work.

caarlos0:
updated to v2, seems to be working fine here

https://github.com/user-attachments/assets/d1eb2552-93a7-4a60-a356-1728b4448bf7

caarlos0:
see #1318

PR #1305, opened 2025-02-04
fix: use CancelIoEx to cancel Windows conInputReader across threads

### LINKED ISSUE #1167: Windows: first character input lost on successive programs
**Describe the bug**

On Windows, if multiple Tea programs are run successively within the same process, the first program runs fine, but subsequent programs lose the first key pressed.  The key is silently eaten, after which point the program appears to operate properly until it terminates.  Then rinse/repeat for subsequent Tea programs.

This does not occur on Linux or Mac.

**Setup**

 - OS: Windows
 - Shell: cmd and Powershell (both affected)
 - Terminal Emulator: Windows Terminal, Windows Powershell Console, Command Prompt
 - Terminal Multiplexer: none

**To Reproduce**

1. Build the example in the source code section for windows
2. Run it
3. Type some keys for the first input prompt ("Test 0") and note that it works fine
4. Type a key for the second input prompt ("Test 1") and note that the first key is eaten, and subsequent keys presses are registered

**Source Code**

```go
package main

import (
	"fmt"

	"github.com/charmbracelet/bubbles/cursor"
	"github.com/charmbracelet/bubbles/textinput"
)

func Input(prompt, value string) {
	ti := textinput.New()
	ti.Prompt = prompt
	ti.Focus()
	p := tea.NewProgram(inputModel{textInput: ti})
	p.Run()
}

type inputModel struct {
	textInput textinput.Model
	quitting  bool
}

func (m inputModel) Init() tea.Cmd {
	return nil
}

func (m inputModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	var cmd tea.Cmd

	switch msg := msg.(type) {
	case tea.KeyMsg:
		switch msg.Type {
		case tea.KeyEnter:
			m.textInput.Cursor.SetMode(cursor.CursorHide)
			m.quitting = true
			return m, tea.Quit
		}
	}
	m.textInput, cmd = m.textInput.Update(msg)
	return m, cmd
}

func (m inputModel) View() string {
	str := m.textInput.View()
	if m.quitting {
		return str + "\n"
	}
	return str
}

func main() {
	for i := range 2 {
		Input(fmt.Sprintf("Test %d: ", i), "default")
	}
}
```

**Expected behavior**

The first key on subsequent Tea programs should not be lost.

**Screenshots**

image

awoodbeck:
The Windows conInputReader uses `windows.CancelIo` to close the read loop's input handle. Windows `CancelIo` docs indicate that call is only appropriate for I/O operations on the same thread as the `CancelIo` call. Since the read loop is in its own goroutine, it's likely to be on another thread from the bubbletea Program.

This change adds a call to Windows `CancelIoEx`, which will cancel I/O operations across threads.

Fixes #1167


iyear:
My local tests have also passed. I am a user of huh, and hope this PR can be merged soon and updated in the relevant library. Thanks!

inputreader_windows.go
```diff
@@ -57,7 +57,7 @@ func newInputReader(r io.Reader) (cancelreader.CancelReader, error) {
 func (r *conInputReader) Cancel() bool {
 	r.setCanceled()
 
-	return windows.CancelIo(r.conin) == nil
+	return windows.CancelIoEx(r.conin, nil) == nil || windows.CancelIo(r.conin) == nil
```
aymanbagabas, line 60: Do we need `windows.CancelIo` anymore?
```suggestion
	return windows.CancelIoEx(r.conin, nil) == nil
```

aymanbagabas (commented):
Could we port this to x/input? This is used in the upcoming v2 release see https://github.com/charmbracelet/x/blob/main/input/cancelreader_windows.go

inputreader_windows.go
awoodbeck: I left the `windows.CancelIo` call in for the very slim chance that the call and the target input handler are on the same thread. I don't seem any harm in keeping it in the code, especially if the `windows.CancelIoEx` call addresses 99% of the use cases. That said, I'll leave the final decision up to you.

aymanbagabas (approved):

aymanbagabas:
Thank you, @awoodbeck! This will be released soon in the next patch release 🙂

PR #1306, opened 2025-02-04
fix: do not swallow context errors

### LINKED ISSUE #1212: context error is swallowed up, preventing errors.Is

andreynering:
- Fixes #1212
- Closes #1217

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -622,7 +622,7 @@ func (p *Program) Run() (Model, error) {
 	model, err := p.eventLoop(model, cmds)
 	killed := p.ctx.Err() != nil || err != nil
 	if killed && err == nil {
-		err = fmt.Errorf("%w: %s", ErrProgramKilled, p.ctx.Err())
+		err = fmt.Errorf("%w: %w", ErrProgramKilled, p.ctx.Err())
 	}
 	if err == nil {
 		// Ensure we rendered the final state of the model.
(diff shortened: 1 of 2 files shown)


andreynering:
So... I just discovered that multiple `%w` are only supported starting on Go 1.20:

```bash
$ env GOTOOLCHAIN=go1.19 go test ./...
go: downloading go1.19 (darwin/arm64)
# github.com/charmbracelet/bubbletea
./tea.go:625:9: fmt.Errorf call has more than one error-wrapping directive %w
FAIL	github.com/charmbracelet/bubbletea [build failed]
FAIL
```

```bash
$ env GOTOOLCHAIN=go1.20 go test ./...
go: downloading go1.20 (darwin/arm64)
ok  	github.com/charmbracelet/bubbletea	0.302
```

By our CI failures, it looks like we're still supposed to support 1.19. I think we can just keep this PR open until we move to Go 1.20 as a minimum requirement.

desertwitch:
Resolved with #1376, #1388

PR #1307, opened 2025-02-04
chore: deprecate some StartupOptions in favor of Cmds

meowgorithm:
This revision depreciates the following `ProgramOption`s in favor of `Cmd`s. All of these will be removed in `v2`.

* `WithAltScreen` (use [`EnterAltScreen`](https://pkg.go.dev/github.com/charmbracelet/bubbletea#EnterAltScreen))
* `WithoutBracketedPaste` (use [`DisableBracketedPaste`](https://pkg.go.dev/github.com/charmbracelet/bubbletea#DisableBracketedPaste))
* `WithMouseCellMotion` (use [`EnableMouseCellMotion`](https://pkg.go.dev/github.com/charmbracelet/bubbletea#EnableMouseCellMotion))
* `WithMouseAllMotion` (use [`EnableMouseAllMotion`](https://pkg.go.dev/github.com/charmbracelet/bubbletea#EnableMouseAllMotion))
* `WithReportFocus` (use [`EnableReportFocus`](https://pkg.go.dev/github.com/charmbracelet/bubbletea#EnableReportFocus))

For equivalent functionality, just return the equivalent `Cmd` in your `Init()`:

```go
func (m Model) Init() tea.Cmd {
    return tea.Batch(tea.EnterAltScreen, tea.EnableMouseAllMotion)
}
```

Additionally, `WithANSICompressor` is deprecated due to performance implications.

PR #1310, opened 2025-02-04
(v2) remove `ProgramOption`s in favor of `Cmd`s`

meowgorithm:
The is following up on API cleanup. For details see: https://github.com/charmbracelet/bubbletea/pull/1307.


aymanbagabas (approved):

meowgorithm:
There's still some work to do here with regard to cleaning up startup options and updating tests.

PR #1314, opened 2025-02-06
feat: add WithWindowSize option to set the initial window size

aymanbagabas:
This adds a new option to set the initial window size of the terminal window. This is useful in testing and required for teatest/v2 after the changes in the renderer. The cell-based cursed renderer heavily relies on the terminal size, and it's important to have a way to set the size manually for tests to run in a non-interactive environment.

### DIFF
--- a/options.go
+++ b/options.go
@@ -295,3 +295,14 @@ func WithColorProfile(profile colorprofile.Profile) ProgramOption {
 		p.profile = profile
 	}
 }
+
+// WithWindowSize sets the initial size of the terminal window. This is useful
+// when you need to set the initial size of the terminal window, for example
+// during testing or when you want to run your program in a non-interactive
+// environment.
+func WithWindowSize(width, height int) ProgramOption {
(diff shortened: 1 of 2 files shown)


raphamorim (approved):
👌

andreynering (approved):

options.go
```diff
@@ -295,3 +295,13 @@ func WithColorProfile(profile colorprofile.Profile) ProgramOption {
 		p.profile = profile
 	}
 }
+
+// WithInitialSize sets the initial size of the terminal window. This is useful
+// when you have a faulty terminal that doesn't report its size correctly, or
+// if you're testing your program in a non-interactive environment.
+func WithInitialSize(width, height int) ProgramOption {
```
aymanbagabas, line 302: Perhaps rename this to `WithWindowSize` to match with WindowSizeMsg 🤔

meowgorithm: I would say so.

PR #1315, opened 2025-02-07
Remove irrelevant option comments

nervo:
Just following https://github.com/charmbracelet/bubbletea/pull/1307 :)

If `With` program options are now deprecated, you should not encourage to use them instead of commands.


meowgorithm:
Great catch: thanks, @nervo!

PR #1318, opened 2025-02-10
feat(v2): update examples

caarlos0:
supercedes #1304


meowgorithm:
The tests are failing here. iirc, @aymanbagabas you changed the way clear screen works, correct? if so, could you update the tests here?

aymanbagabas (approved):

aymanbagabas:
> The tests are failing here. iirc, @aymanbagabas you changed the way clear screen works, correct? if so, could you update the tests here?

Correct, fwiw, I'm making a few changes to the renderer that will affect the tests, we will need to update the tests again after that.

aymanbagabas:
I've updated the testdata and golden files in `v2-exp` to use the latest renderer, which includes a behavioral breaking change. Now, on program exit, the renderer will always move the cursor to the beginning of the next line of the screen, scrolling the terminal if necessary. This means examples and programs that include a trailing newline in their `View()` won't need to do so anymore.

aymanbagabas (approved):

.github/workflows/build.yml
```diff
@@ -4,9 +4,6 @@ on: [push, pull_request]
 jobs:
   build:
     uses: charmbracelet/meta/.github/workflows/build.yml@main
-
-  build-go-mod:
```
aymanbagabas, line 8: This uses the go.mod version to test the package while the `build` job uses the latest stable go version

caarlos0: ah, right. will undo

aymanbagabas (approved):

PR #1321, opened 2025-02-10
fix: reset renderer and cursor when releasing terminal

aymanbagabas:
The cursed renderer moves the cursor between renders, which can cause issues when the terminal is released and restored. For instance, when we exec a command using tea.Exec, we want the cursor position to remain the same after the command finishes and goes back to the same position as before the command was executed.

With tea.Suspend on the other hand, we want to make sure we reset the cursor position because the program stops executing and is put to suspend. Resuming the program again initializes the renderer and its cursor position again from scratch as if it was a new program.

This wasn't an issue for the standard renderer because we always put the cursor at the lower left corner of the screen. Releasing and restoring the program wasn't an issue because the cursor was always at the same position.

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -1050,6 +1050,10 @@ func (p *Program) recoverFromPanic() {
 // ReleaseTerminal restores the original terminal state and cancels the input
 // reader. You can return control to the Program with RestoreTerminal.
 func (p *Program) ReleaseTerminal() error {
+	return p.releaseTerminal(false)
+}
+
+func (p *Program) releaseTerminal(reset bool) error {
 	atomic.StoreUint32(&p.ignoreSignals, 1)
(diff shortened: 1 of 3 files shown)


tea.go
```diff
@@ -1050,6 +1050,10 @@ func (p *Program) recoverFromPanic() {
 // ReleaseTerminal restores the original terminal state and cancels the input
 // reader. You can return control to the Program with RestoreTerminal.
 func (p *Program) ReleaseTerminal() error {
+	return p.releaseTerminal(false)
```
aymanbagabas, line 1053: I'm not sure if this is the right place to have the reset flag. Perhaps we should change the signature of `p.RestoreTerminal(reset bool) error`

caarlos0 (approved):

Issue #1322, opened 2025-02-10
bug: last rendered line gets ignored


aymanbagabas:
Taken from https://github.com/charmbracelet/bubbletea/discussions/1317

I recently upgraded from bubbletea from v0.26.5 to v1.3.0 in my project: https://github.com/antonmedv/walk.

<img width="1098" alt="image" src="https://github.com/user-attachments/assets/98ada3ca-bde5-4016-b6ed-fe56139d7535" />

And I encountered some very strange behavior with rendering. You can try it yourself on walk version 1.11.0. 

- yank some file path [y]
- see bar [copied: some path]
- move cursor with arrows
- last line disappeared! 👻

I was struggling to find a bug in my code, but after downgrading - everything is fine again.

<img width="882" alt="Shottr 2025-02-08 19 49 14" src="https://github.com/user-attachments/assets/1a25f04d-92e1-4d5d-ba32-1b173800a246" />
<img width="882" alt="Терминал 2025-02-08 19 49 46" src="https://github.com/user-attachments/assets/c1c027be-9a0e-4fb3-b542-deb996232dcb" />

Does somebody experience similar problems?

PR #1323, opened 2025-02-10
fix: renderer: use newline instead of cud1 to move cursor down

aymanbagabas:
Using cud1 (`\x1b[B`) won't scroll the terminal if the cursor is already at the bottom of the screen. Instead, use a newline character to move the cursor down and scroll the terminal if necessary.

Fixes: https://github.com/charmbracelet/bubbletea/issues/1322


antonmedv:
Awesome! Cool!

PR #1327, opened 2025-02-11
chore(dependabot): run weekly on monday

andreynering:


.github/dependabot.yml
```diff
@@ -1,36 +1,52 @@
 version: 2
+
 updates:
   - package-ecosystem: "gomod"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+      day: "monday"
+      time: "05:00"
```
aymanbagabas, line 9: I would make it earlier in the morning so that we get a chance to talk about them in the meetings in case of any issues

aymanbagabas (commented):
We should do something like [lint-sync](https://github.com/charmbracelet/meta/blob/main/.github/workflows/lint-sync.yml) to synchronize the dependabot files across all of our repositories

.github/dependabot.yml
andreynering: 5am looked soon enough to me. What time do you suggest?

aymanbagabas: Oh my bad, I use the AM/PM system 😅

Issue #1313, opened 2025-02-05
[Windows] Can't select text on the terminal when inside a program


rajayonin:
**Describe the bug**
When inside a Bubble Tea program, I can't select any text on my Windows Terminal. It doesn't work on either PowerShell or CMD, but it does on WSL.

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS: Windows
 - Shell: PowerShell/CMD
 - Terminal Emulator: Windows Terminal
 - Terminal Multiplexer: None

**To Reproduce**
1. Compile and run a Bubble Tea program
2. Try to click and drag on the text
3. Nothing happens

**Source Code**
I've used the code from the ['Credit Card Form' example](https://github.com/charmbracelet/bubbletea/blob/main/examples/credit-card-form/main.go).

<details>

<summary>Full code</summary>

```go
package main

import (
	"fmt"
	"log"
	"strconv"
	"strings"

	"github.com/charmbracelet/bubbles/textinput"
	tea "github.com/charmbracelet/bubbletea"
	"github.com/charmbracelet/lipgloss"
)

func main() {
	p := tea.NewProgram(initialModel())

	if _, err := p.Run(); err != nil {
		log.Fatal(err)
	}
}

type (
	errMsg error
)

const (
	ccn = iota
	exp
	cvv
)

const (
	hotPink  = lipgloss.Color("#FF06B7")
	darkGray = lipgloss.Color("#767676")
)

var (
	inputStyle    = lipgloss.NewStyle().Foreground(hotPink)
	continueStyle = lipgloss.NewStyle().Foreground(darkGray)
)

type model struct {
	inputs  []textinput.Model
	focused int
	err     error
}

// Validator functions to ensure valid input
func ccnValidator(s string) error {
	// Credit Card Number should a string less than 20 digits
	// It should include 16 integers and 3 spaces
	if len(s) > 16+3 {
		return fmt.Errorf("CCN is too long")
	}

	if len(s) == 0 || len(s)%5 != 0 && (s[len(s)-1] < '0' || s[len(s)-1] > '9') {
		return fmt.Errorf("CCN is invalid")
	}

	// The last digit should be a number unless it is a multiple of 4 in which
	// case it should be a space
	if len(s)%5 == 0 && s[len(s)-1] != ' ' {
		return fmt.Errorf("CCN must separate groups with spaces")
	}

	// The remaining digits should be integers
	c := strings.ReplaceAll(s, " ", "")
[...]

PR #1340, opened 2025-02-24
fix: windows: enable mouse mode on demand

aymanbagabas:
Since we added native Windows API input events support, we always enable mouse events even if the user doesn't want them. This is a problem because it breaks applications that don't expect mouse events and selection stops working without any modifier key.

Fixes: https://github.com/charmbracelet/bubbletea/issues/1313


inputreader_other.go
```diff
@@ -9,6 +9,6 @@ import (
 	"github.com/muesli/cancelreader"
 )
 
-func newInputReader(r io.Reader) (cancelreader.CancelReader, error) {
+func newInputReader(r io.Reader, _ bool) (cancelreader.CancelReader, error) {
```
raphamorim, line 12: we don't care about that right? It's just the new signature I assume

inputreader_windows.go
```diff
@@ -38,11 +38,16 @@ func newInputReader(r io.Reader) (cancelreader.CancelReader, error) {
...
 
-	originalMode, err := prepareConsole(conin,
-		windows.ENABLE_MOUSE_INPUT,
+	modes := []uint32{
 		windows.ENABLE_WINDOW_INPUT,
 		windows.ENABLE_EXTENDED_FLAGS,
-	)
+	}
+
+	if enableMouse {
```
raphamorim, line 46: I think would be good to add why enableMouse isn't default, other I feel tends to move back modes initilization.

What do you think?

aymanbagabas: added a comment

raphamorim: thank you!

raphamorim (approved):

andreynering (approved):

PR #1347, opened 2025-03-07
Ensure nils are removed from sequences

SteveNewson:
When a sequence is used, nils trigger a go routine in `tea.go` here:
```
case sequenceMsg:
  go func() {
    // Execute commands one at a time, in order.
    for _, cmd := range msg {
      if cmd == nil {
        continue
      }
      //...
    }
  }()
```
Even though the go routine is short-lived, this can cause a significant performance hit to the runtime. The Batch mechanism eliminates nils before they are processed. This change replicates that behaviour in `sequenceMsg`.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -13,6 +13,25 @@ import (
 //		       return tea.Batch(someCommand, someOtherCommand)
 //	    }
 func Batch(cmds ...Cmd) Cmd {
+	return trimNilCommands[BatchMsg](cmds...)
+}
+
+// BatchMsg is a message used to perform a bunch of commands concurrently with
+// no ordering guarantees. You can send a BatchMsg with Batch.
+type BatchMsg []Cmd
+
+// Sequence runs the given commands one at a time, in order. Contrast this with
+// Batch, which runs commands concurrently.
+func Sequence(cmds ...Cmd) Cmd {
+	return trimNilCommands[sequenceMsg](cmds...)
+}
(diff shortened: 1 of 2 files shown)


SteveNewson:
This issue hit me harder because I was using a sequence in an `Update()` method where a batch would have been more appropriate. Since the majority of updates resulted in no new commands, I was generating a lot of sequences with nils in them. This resulted in an extremely laggy UI.

meowgorithm:
Thanks for the PR and apologies for the delay on this one. This generally makes sense to me and I think its a good one to get in, we just need to give it a proper look, which we intend to.

aymanbagabas (approved):
LGTM, thank you @SteveNewson!

meowgorithm:
@SteveNewson That’s a hot type constraint. Well done.

@aymanbagabas Do you know why BatchMsg is exposed? I’d think it shouldn’t be, and if so maybe we unexpose it in v2z

aymanbagabas:
> @aymanbagabas Do you know why BatchMsg is exposed? I’d think it shouldn’t be, and if so maybe we unexpose it in v2.

I don't know why but it's been there for a long time. Perhaps unexpose it in v2 is the right time.

aymanbagabas:
Superseded by https://github.com/charmbracelet/bubbletea/pull/958. Thank you @SteveNewson!

PR #1355, opened 2025-03-17
Fix: Windows: first character input lost on successive programs

joshallenit:
Explanation here: https://github.com/charmbracelet/bubbletea/issues/1167#issuecomment-2728251993

Note: I don't think it is possible to unit test this in a meaningful way because to reproduce the issue requires using console input, but please correct me if I'm wrong.

### DIFF
--- a/inputreader_windows.go
+++ b/inputreader_windows.go
@@ -67,6 +67,8 @@ func newInputReader(r io.Reader, enableMouse bool) (cancelreader.CancelReader, e
 func (r *conInputReader) Cancel() bool {
 	r.setCanceled()
 
+	// Warning: These cancel methods do not reliably work on console input
+	// 			and should not be counted on.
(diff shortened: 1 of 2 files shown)


joshallenit:
Looks like main branch is broken as all of the PR's are failing for the same reason. `go mod tidy` needs to be called, although that might not be the only reason. When I run `go mod tidy` all it does it update the go version, which I don't think is desired, at least not in this PR. Can a code owner fix the main branch?

meowgorithm:
Hey there! Thanks for flagging this; we're looking into it right now.

meowgorithm:
Alright, the build's fixed on `main`. We're also working on the lint stuff now as well.

andreynering:
@joshallenit The CI is green again. You can rebase your branch with `main`.

PR #1369, opened 2025-03-25
feat(v2): merge main into v2-exp

caarlos0:
refs https://github.com/charmbracelet/bubbles/pull/756


meowgorithm (approved):

key_windows.go
```diff
@@ -0,0 +1,360 @@
+//go:build windows
```
caarlos0, line 1: @aymanbagabas do we need this in v2?

aymanbagabas: No, we don't. This logic is part of x/input

PR #1398, opened 2025-04-22
fix: always enable basic keyboard enhancements to disambiguate keys

aymanbagabas:
Right now, there is inconsistencies between Windows and other *nix platforms when it comes to keyboard enhancement handling. Since we use the Windows Console API on Windows, we always get disambiguated keys and release events.

To make the experience consistent across platforms, we always enable disambiguated keys for *nix platforms, and we ignore release events on Windows unless they are explicitly requested, the same way we do for *nix platforms.


caarlos0 (approved):

bashbunni:
Ah ok, so the key release stuff wouldn't be enabled by default anymore when you allow keyboard enhancements? It would be opt-in for Windows too?

PR #1418, opened 2025-05-14
chore(v2): pin the right lipgloss v2 tag on `go.mod`

### LINKED ISSUE #1417: fix(deps): Lip Gloss v2 in examples
examples is using an invalid version for lip gloss v2, introduced in #1401 should be https://github.com/charmbracelet/lipgloss/releases/tag/v2.0.0-beta.1

andreynering:
Fixes #1417


bashbunni (changes requested):
ah we need to add the charmtone and glamour deps back in, they're giving me (and CI) an error

andreynering:
Fixed. I just couldn't pin the tag for now, because Glamour depends on changed from the `v2-exp` branch:

https://github.com/charmbracelet/glamour/blob/eeeced3bb3c6bf4b2a55c66ede53115dfd642d4d/go.mod#L9

There is another error on CI, but that was already happening before.

PR #1422, opened 2025-05-24
docs(examples/textinput): support virtual cursor

meowgorithm:
This adds virtual cursor support to the `textinput` example.

cc: @kujtimiihoxha


kujtimiihoxha:
I think I tried this here and the textarea if you turn it on it will not blink.

meowgorithm:
@kujtimiihoxha Good catch: that'll be fixed here: https://github.com/charmbracelet/bubbles/pull/794

PR #1426, opened 2025-05-26
V2 viewable

meowgorithm:
This replaces `CursorView` (and potentially `ViewModel`) with a simpler, more extensible API.

```go
type Viewable interface {
    View() View
}
```

In practice, it looks like:

```go
// A simple case where the view is just a string.
func (m Model) View() tea.View {
    // ...
    return tea.NewView(str)
}

// A more complex case where the view is a string with
// cursor and background color metadata.
func (m Model) View() tea.View {
    // ...
    v := tea.NewView(str)
    v.SetCursor(cur)
    v.SetBackgroundColor(c)
    return v
}
```

Properties like the cursor and background color are optional, and `View` could be extended down the road to include other metadata, like `lipgloss.Canvas`. Note that the background color _should_ be attached to the view to avoid race conditions like we saw with `tea.SetCursorPosition`.

It would be my suggestion _not_ to merge `Viewable` with `Model` in order to eventually support view alternatives like the following:

```go
func (m Model) View(b *screen.Buffer)
```

### DIFF
--- a/examples/textinput/main.go
+++ b/examples/textinput/main.go
@@ -70,7 +70,9 @@ func (m model) View() tea.View {
 		str += "\n"
 	}
 
-	return tea.NewView(str).Cursor(c)
+	v := tea.NewView(str)
+	v.SetCursor(c)
+	return v
 }
 
 func (m model) headerView() string { return "What’s your favorite Pokémon?\n" }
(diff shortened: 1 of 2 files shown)


meowgorithm:
v2 is out! Closing accordingly.

PR #1429, opened 2025-05-29
Ensure rendering before exec

raphaelvigee:
### Describe your changes
I'm building an application that has a pretty strict requirement that the view has been rendered before diving into exec

### Checklist before requesting a review

- [x] I have read `CONTRIBUTING.md`
- [x] I have performed a self-review of my code

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -99,7 +99,9 @@ func (c *osExecCommand) SetStderr(w io.Writer) {
 }
 
 // exec runs an ExecCommand and delivers the results to the program as a Msg.
-func (p *Program) exec(c ExecCommand, fn ExecCallback) {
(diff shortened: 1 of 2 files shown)


raphaelvigee:
@aymanbagabas @meowgorithm Whats the path towards getting that merged ?

aymanbagabas:
Hi @raphaelvigee, could you provide an example of how would this look like and some test cases?

raphaelvigee:
It doesnt change anything functionally, and i dont really know what to test for, its just that in the current implementation, because the renderer is ticker based, bubbletea could enter the exec mode before the UI had time to enter "exec mode" (in my case clearing the screen), this ensure the renderering is consistent with the expected state

PR #1446, opened 2025-07-03
feat: create a panic trace file

aymanbagabas:
When Bubble Tea panics, it is really useful to have a trace file that contains the panic message and the stack trace. This does that exactly.

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -19,6 +19,7 @@ import (
 	"os/signal"
 	"runtime/debug"
 	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"syscall"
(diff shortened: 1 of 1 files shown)


taigrr:
thank you! i've been trying to catch a crash from a panic by zooming my screen down to 3pt font.

tea.go
```diff
@@ -1131,8 +1132,19 @@ func (p *Program) recoverFromPanic(r any) {
...
+	fmt.Fprintf(os.Stderr, "Caught panic:\r\n\r\n%s\r\n\r\nRestoring terminal...\r\n\r\n", rec)
+	stack := strings.ReplaceAll(fmt.Sprintf("%s\n", debug.Stack()), "\n", "\r\n")
+	fmt.Fprint(os.Stderr, stack)
+	f, err := os.Create(fmt.Sprintf("bubbletea-panic-%d.log", time.Now().Unix()))
+	if err == nil {
+		defer f.Close()        //nolint:errcheck
+		fmt.Fprintln(f, rec)   //nolint:errcheck
+		fmt.Fprintln(f)        //nolint:errcheck
+		fmt.Fprintln(f, stack) //nolint:errcheck
+	}
```
andreynering, line 1147: Question: do we want to have this enabled by default?

An easy option to make this configurable is to make it opt-in or opt-out via an env, so it can be set by the caller or the app itself:

```go
func init() {
    os.Setenv("BUBBLETEA_PANIC_TRACE_FILE", "1")
    // or
    os.Setenv("BUBBLETEA_DISABLE_PANIC_TRACE_FILE", "1")
}
```

andreynering (approved):

tea.go
meowgorithm: I'd make this opt in by default so Bubble Tea applications don't litter directories with panic notices. This behavior could be overridden with a `ProgramOption`.

caarlos0 (approved):

tea.go
aymanbagabas: @andreynering I've made it opt-in via the `TEA_DEBUG` environment variable.

PR #1473, opened 2025-08-15
fix: send commands in a goroutine without executing them immediately

aymanbagabas:
The commands channel handler handles commands and their execution. We should not execute them ourself in the event loop and instead send them to the commands channel where the handles will take care of executing them.

Fixes: f5da8d068af7 (fix: handle nested SequenceMsg in event loop and use sync.WaitGroup f… (#1463))


tea.go
```diff
@@ -492,23 +503,16 @@ func (p *Program) eventLoop(model Model, cmds chan Cmd) (Model, error) {
...
-									p.Send(cmd())
+									p.sendCmd(cmds, cmd)
 								}()
 							}
 							wg.Wait()
 						case sequenceMsg:
 							for _, cmd := range msg {
-								if cmd == nil {
-									continue
-								}
```
caarlos0, line 510: were these checks needed? if so, might want to add them to sendCmd

caarlos0 (approved):

PR #1499, opened 2025-09-30
feat: add support for terminal progress bars

aymanbagabas:
This adds support for displaying progress bars in terminals that supports it such as Windows Terminal and Ghostty.

See https://learn.microsoft.com/en-us/windows/terminal/tutorials/progress-bar-sequences for more information.

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -154,6 +154,17 @@ const (
 	ProgressBarWarning
 )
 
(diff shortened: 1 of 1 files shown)


caarlos0 (approved):

meowgorithm:
Eh, I guess we also need to fix lint.

PR #1503, opened 2025-10-01
fix: race condition

caarlos0:
this fixes a race condition in `outputBuf`.


aymanbagabas:
Thanks @caarlos0! Superseded by https://github.com/charmbracelet/bubbletea/pull/1502 as we are not really benefiting from the RWMutex, both methods mutate the buffer.

PR #1509, opened 2025-10-09
added forbidigo linter and allow two printf calls

dennyreeh:


lrstanley:
I believe changes to the golangci-lint configs should be done on [this repo](https://github.com/charmbracelet/meta/blob/main/golangci.yml), as there are Actions in this repo that sync from that repo. If changed here, they would get overwritten.

christiandins:
Hi lrstanley,
thanks for the hint. I've created a pull request in that repo too.
https://github.com/charmbracelet/meta/pull/247

Issue #304, opened 2022-04-26
Force newline after rendering the final View


erikgeiser:
If the final call to `View()` (after emitting `tea.Quit`) does not end with a newline it is immediately overwritten by the shell prompt or by any other output. This can be a bit surprising and it's hard to debug if you don't know what happens.

It can be avoided by appending a newline to the result of the final call to `View()` if this result is not empty.

meowgorithm:
Making this change will slightly alter the final output of some programs, however it will remove a quirk which we've seen confuse new users.

The former behavior can basically be recreated via some additional logic in `View()`, which is where such rendering logic should be anyway.

* * *

Here are some illustrations of the current behavior:

Normal view. The help text resides on the final line.
```
[x] UI
[ ] More UI
Some help text
````

After quit. The developer is exploiting a quirk in which the final view doesn't contain a newline, so the help text prints over it on exit.
```
[x] UI
[ ] More UI
user@machine:~ $: _
```

If the prompt it shorter than the TUI's final line some artifacts from the final line will appear after the prompt.

```
[x] UI
[ ] More UI
$: _ Help Text
```

praagyajoshi:
As a new user, I just got bit by this behaviour! And I have to admit, it really stumped me for quite a few hours where I spent figuring out whether:
* there was an issue an my app logic
* there was an issue in lipgloss (which is what I was using to render the items)

I eventually figured out an extra newline was needed and submitted an issue here: https://github.com/charmbracelet/bubbletea/issues/827.

I will definitely recommend updating the default behaviour (to print an extra newline) with an option to turn it off if needed.

virtuallyunknown:
> As a new user, I just got bit by this behaviour! And I have to admit, it really stumped me for quite a few hours where I spent figuring out whether:

As a new user, and new to go in general I also assumed this is a mistake on my end, so same story as you. Would be nice if there was a note about this folks.

CalebJohnHunt:
Is there still interest in this? I've made a [proof of concept](https://github.com/CalebJohnHunt/bubbletea/commit/f61cfa1173c31fe5e8ed26efd508badaeca408e2), but I'm open to changing the implementation.

PR #1514, opened 2025-10-15
fix: renderer: add a trailing newline when exiting inline mode

aymanbagabas:
When exiting inline mode, we need to insert a trailing newline to ensure that the terminal prompt appears on a new line. This change adds a flag to track whether we are in inline mode and modifies the close method to append a newline if necessary.

Fixes: https://github.com/charmbracelet/bubbletea/issues/304


aymanbagabas:
After discussing this with @meowgorithm and looking at the use cases, we decided that we don't need this because it can break programs that want to completely erase the view on exit and adding a trailing newline breaks that use case.
Models can still add a trailing newline on exit to preserve the program's last line.

PR #1532, opened 2025-11-03
perft: lazy render model after event loop

aymanbagabas:
On a large program, this can reduce repeated model render calls when the model is getting updated rapidly. This change makes the model stored as an atomic pointer for safe concurrent access between the event loop and the render loop.

The model is now stored internally as an atomic pointer, and updated after each event loop iteration. The render loop reads the latest model from the atomic pointer before rendering its View.

Benchmarks show a slight reduction in allocations for the TeaTest example.
```
goos: darwin
goarch: arm64
pkg: github.com/charmbracelet/bubbletea/v2
cpu: Apple M3 Max
          │   old.txt   │            new.txt            │
          │   sec/op    │   sec/op     vs base          │
TeaRun-16   321.2µ ± 3%   318.3µ ± 4%  ~ (p=0.247 n=10)

          │   old.txt    │            new.txt             │
          │     B/op     │     B/op      vs base          │
TeaRun-16   623.0Ki ± 0%   622.2Ki ± 0%  ~ (p=0.105 n=10)

          │   old.txt   │              new.txt               │
          │  allocs/op  │  allocs/op   vs base               │
TeaRun-16   1.837k ± 0%   1.828k ± 0%  -0.52% (p=0.000 n=10)
```

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -394,9 +394,9 @@ type Program struct {
 	// be capped at 120.
 	fps int
 
-	// initialModel is the initial model for the program and is the only
-	// required field when creating a new program.
-	initialModel Model
+	// model stores the last updated model as an atomic pointer for safe
+	// concurrent access.
+	model Model
 
 	// disableRenderer prevents the program from rendering to the terminal.
 	// This can be useful for running daemon-like programs that don't require a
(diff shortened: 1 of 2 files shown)


aymanbagabas:
Actually, this might not be a good idea because it will significantly slow down since it synchronizes the event and render loop which we don't want here :/
Closing...

PR #1534, opened 2025-11-05
refactor: keyboard enhancements api improvements

aymanbagabas:
This change improves clarity and usability of the keyboard enhancements API by introducing a dedicated struct to encapsulate keyboard enhancement features. Now, key disambiguation is always requested and cannot be opted out of, simplifying the API. Terminals should ignore the request if they do not support keyboard enhancements.

### DIFF
--- a/keyboard.go
+++ b/keyboard.go
@@ -9,11 +9,25 @@ import (
 type KeyboardEnhancementsMsg struct {
 	// Flags is a bitmask of supported keyboard enhancement features.
 	// See [ansi.KittyReportEventTypes] and other constants for details.
+	//
+	// Example:
+	//
+	//  ```go
+	//  // The hard way
+	//  if msg.Flags&ansi.KittyReportEventTypes != 0 {
+	//     // Terminal supports reporting different key event types
+	//  }
+	//
+	//  // The easy way
+	//  if msg.SupportsEventTypes() {
+	//     // Terminal supports reporting different key event types
+	//  }
+	//  ```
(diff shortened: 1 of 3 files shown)


keyboard.go
```diff
@@ -7,25 +7,13 @@ import (
...
-// disambiguous keys as escape codes.
-func (k KeyboardEnhancementsMsg) SupportsKeyDisambiguation() bool {
-	return k.Flags&ansi.KittyDisambiguateEscapeCodes != 0
-}
-
-// SupportsKeyReleases returns whether the terminal supports key release
-// events.
-func (k KeyboardEnhancementsMsg) SupportsKeyReleases() bool {
+// SupportsEventTypes returns whether the terminal supports reporting
+// different types of key events (press, release, repeat).
```
meowgorithm:
```suggestion
// SupportsEventTypes returns whether the terminal supports reporting
// different types of key events (press, release, and repeat).
```

keyboard.go
```diff
@@ -7,25 +7,13 @@ import (
 // KeyboardEnhancementsMsg is a message that gets sent when the terminal
 // supports keyboard enhancements.
 type KeyboardEnhancementsMsg struct {
+	// Flags is a bitmask of supported keyboard enhancement features.
+	// See [ansi.KittyReportEventTypes] and other constants for details.
```
meowgorithm: Does this need to be user-facing? If so, show an example of how to use it.

PR #1544, opened 2025-11-16
feat: allow chainable msg filters; add MouseThrottleFilter

lrstanley:
## Changes in this PR

Made some tweaks to message filtering, as I think this will help others. If y'all don't think it will be too useful, or would rather approach it a different way, feel free to close this out.

- Converted `WithFilter(filter)` to `WithFilters(filters...)` to allow chainable filters, with a `MsgFilter` type to clean things up a little.
- Added an out-of-box `MouseThrottleFilter` filter, to cover throttling of high-frequency mouse messages. This logic is also used in crush [here](https://github.com/charmbracelet/crush/blob/0f39613d53dbb7b0a42d2b2d5541e63a9369f555/internal/tui/tui.go#L41-L54), and one of my apps, and I suspect such functionality will help others improve performance when they don't need the high-frequency tracking.

---

### DIFF
--- a/examples/prevent-quit/main.go
+++ b/examples/prevent-quit/main.go
@@ -20,14 +20,14 @@ var (
 )
 
 func main() {
-	p := tea.NewProgram(initialModel(), tea.WithFilter(filter))
+	p := tea.NewProgram(initialModel(), tea.WithFilters(preventUnsavedFilter))
 
 	if _, err := p.Run(); err != nil {
 		log.Fatal(err)
 	}
 }
 
-func filter(teaModel tea.Model, msg tea.Msg) tea.Msg {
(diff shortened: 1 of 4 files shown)


lrstanley:
Tests currently breaking due to https://github.com/charmbracelet/bubbletea/blob/f60735dff10c208b3398dde25a39b1d27eb6d266/examples/go.mod#L7

aymanbagabas:
> Tests currently breaking due to
> 
> https://github.com/charmbracelet/bubbletea/blob/f60735dff10c208b3398dde25a39b1d27eb6d266/examples/go.mod#L7

I removed this in the latest v2-exp commit

aymanbagabas:
Thank you so much for your contribution here @lrstanley!

We accidentally closed your PR after merging v2-exp. If you still think this is relevant, please re-open your changes and, we will take a look as soon as possible.

PR #1551, opened 2025-11-24
v2: fix(renderer): debounce resize renders

aymanbagabas:
This change introduces a debounce mechanism to the cursed renderer's resize handling. When a resize event occurs, the renderer will now wait for a short duration before processing the resize. This prevents multiple full resize renders from occurring in quick succession, which can lead to performance issues and visual artifacts.


aymanbagabas:
I don't know if we need this tbh 🤔

PR #1555, opened 2025-12-04
Add WithoutOutput option for input-only mode

robin-samuel:
Add new WithoutOutput() option that disables visual output while preserving raw mode for proper keyboard input handling. This allows arrow keys and other special keys to work without requiring Enter, while standard output functions like println() and logging libraries work correctly.

This solves a common issue where users want Bubbletea's input handling but need to use standard output for logging or debugging without the renderer interfering with cursor positioning.

Changes:
- Add WithoutOutput() option in options.go
- Add disableOutput flag to Program struct
- Use term.MakeRawOutput() when disableOutput is enabled
- Add counter example demonstrating the usage
- Update go.mod with local x/term replace directive for development

The implementation uses the new term.MakeRawOutput() function which sets raw mode for input while preserving output processing flags (OPOST/ONLCR), ensuring that newlines are automatically converted to CR+LF as expected by standard output functions.

### DIFF
--- a/tty_windows.go
+++ b/tty_windows.go
@@ -34,7 +34,7 @@ func (p *Program) initInput() (err error) {
 
 	// Save output screen buffer state and enable VT processing.
 	// Skip output modifications in disableOutput mode to preserve normal output behavior.
-	if f, ok := p.output.(term.File); ok && term.IsTerminal(f.Fd()) && !p.disableOutput {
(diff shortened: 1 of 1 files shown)


robin-samuel:
**Note:** This PR depends on charmbracelet/x#687 (the `MakeRawOutput()` function). 

The `go.mod` currently has a temporary replace directive for local development. Once the x/term PR is merged, I'll update this to use the proper version and remove the replace directive.

PR #1575, opened 2026-01-26
feat(render): add option to show render time in the view

aymanbagabas:
This change makes it possible to display the time taken to render each frame on the terminal screen itself. This is controlled by the TEA_RENDER_DEBUG environment variable. When set to true, the renderer will overlay the render time on the view.


andreynering (approved):
Very cool

PR #1576, opened 2026-02-04
fix(render): execute insert above immediately

aymanbagabas:
This commit modifies the behavior of the `insertAbove` method in the renderer to execute the insertion of lines immediately rather than queuing them for later processing. This change ensures that lines are inserted above the current view without delay.


meowgorithm:
Just for some more context here, the problem this solves for is that `tea.Print` family statements currently return instantly, sending a `Msg` to the messages queue (channel) requesting a print, so in cases like the following, order is not guaranteed:

```go
tea.Sequence(tea.Println("first"), tea.Raw("\r\nsecond\r\n"))
```

In the above case, the output will almost always result in the following, because `tea.Raw` does print immediately:

```
second
first
```

PR #1579, opened 2026-02-07
fix(renderer): add modifyOtherKeys support for tmux

erikstmartin:
Modified keys (shift+enter, ctrl+i, etc.) don't work in tmux because tmux doesn't support the Kitty keyboard protocol. tmux requires applications to request extended-keys using modifyOtherKeys (CSI > 4 ; 2 m).

The workaround is setting 'extended-keys always' in tmux config (tmux 3.2+), but users may not know this is the issue, and forcing extended-keys for all applications may not be desired. This fix sends modifyOtherKeys requests alongside Kitty protocol sequences, to enable proper key handling in tmux automatically.

- [x] I have created a discussion that was approved by a maintainer (for new features).


aymanbagabas:
@erikstmartin It looks like we need both `ansi.SetModifyOtherKeys1` and `ansi.SetModifyOtherKeys2` for shift+enter, ctrl+i, ctrl+c, and others to work :thinking:

You could try these keys with
```sh
printf '\x1b[>4;1m' && cat -e
printf '\x1b[>4;2m' && cat -e
printf '\x1b[>4;1m\x1b[>4;2m' && cat -e
```

erikstmartin:
@aymanbagabas you shouldn't need both. ansi.SetModifyOtherKeys1 would only encode modified keys when they don't have ASCII equivelants. ansi.SetModifyOtherKeys2 encodes all modified keys, even if they have ASCII equivelants.   So adding it wouldn't change any behavior, ansi.SetModifyOtherKeys2 would take precedence.

aymanbagabas:
I see. I think we need `ansi.SetModifyOtherKeys2` to get things like <kbd>shift+enter</kbd>` to work

Issue #1595, opened 2026-02-25
v2: Missing WindowSizeMsg on windows systems


robin-samuel:
**Describe the bug**
`WindowSizeMsg` doesnt get send anymore in v2.0.0 on window resizes. Only on `tea.RequestWindowSize`. Only on Windows. Mac and Linux seem fine.

**Setup**
 - Windows 11
 - Windows Powershell, Windows Powershell 7, CMD
 - Windows Consolehost, Windows Terminal

**To Reproduce**
1. Run `go run .\examples\window-size\main.go`
2. Resize the terminal window
3. Nothing happens

**Source Code**
Not needed bubbletea has an example

**Expected behavior**
It should show window size msgs on resize but nothing appears.

**Screenshots**
This was the behavior in v2.0.0 RC 2 (as it should be)

https://github.com/user-attachments/assets/ef6f34fd-b8b8-4100-9fa9-35669c18f34c

 This is the current behavior in v2.0.0 (broken)

https://github.com/user-attachments/assets/630d32b4-e61a-4369-b859-b9d6ad058535

**Additional context**
I found that the commit b3661ce3d63f003545323b721212562c28647e0f introduced tty as default input seems like this doesnt work for windows. Using `tea.WithInput(os.Stdin)` as option fixes this.

aymanbagabas:
Hi @robin-samuel, thanks again for reporting this one. It should be resolved in the latest [v2.0.1 release](https://github.com/charmbracelet/bubbletea/releases/tag/v2.0.1) 🙂

PR #1596, opened 2026-02-25
fix: check if os.Stdin is a terminal before opening the TTY

aymanbagabas:
This change checks if os.Stdin is a terminal before attempting to open the TTY for input. If os.Stdin is not a terminal, it will proceed to open the TTY as before.

Fixes: https://github.com/charmbracelet/bubbletea/issues/1595


robin-samuel:
Fixes the issue for me!

PR #1633, opened 2026-03-18
fix: removed string concat in a loop is inefficient warning

corbinlazarone:
# Summary
New users following the tutorial get "using string += string in a loop is inefficient" from the go lsp. I removed this warning by replacing the string concatenation with strings.builder go package.


themavik (commented):
Builder change is fine; this could still trip staticcheck if other examples/tutorials concatenate in a loop the same way, so worth grepping for the same pattern while you are here.

PR #1640, opened 2026-03-20
docs: fix progress download comment grammar

Rohan5commit:
## Summary
- fix a grammar issue in the `examples/progress-download` comment
- keep the change limited to one file with no behavior impact

## Related issue
- N/A

## Guideline alignment
- Minimal comment-only change aligned with the repo contribution guidance: https://github.com/charmbracelet/bubbletea/contribute

## Validation/testing
- Not run; comment-only change

### DIFF
--- a/examples/progress-download/main.go
+++ b/examples/progress-download/main.go
@@ -66,8 +66,8 @@ func main() {
 	}
 	defer resp.Body.Close() // nolint:errcheck
 
-	// Don't add TUI if the header doesn't include content size
-	// it's impossible see progress without total
+		// Don't add TUI if the header doesn't include content size, since it's
+		// impossible to show progress without a total.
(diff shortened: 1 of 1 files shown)


MukundaKatta:
Grammar tweak reads better — one small thing: the new lines are indented with **two tabs** where the original had one, but the surrounding \`if\` block is indented with a single tab, so the comment is now indented one level deeper than the code it documents. \`gofmt\` or \`go vet\` should flag it, and reviewers tend to squint at this.

Current diff (tabs shown as `⭾`):

```
⭾⭾// Don't add TUI if the header doesn't include content size, since it's
⭾⭾// impossible to show progress without a total.
⭾if resp.ContentLength <= 0 {
```

Dropping one tab from each comment line should line it up with the \`if\` below.

Issue #761, opened 2023-06-13
"open /dev/tty: no such device or address" when running in Docker with no tty


jonbretman:
**Describe the bug**
It seems that when running inside a Docker container with no tty `Run()` returns an error:

```
open /dev/tty: no such device or address
```

**Setup**
Please complete the following information along with version numbers, if applicable.
 - Ubuntu (docker)

**To Reproduce**
Simple bubbletea program:
```go
package main

import (
	tea "github.com/charmbracelet/bubbletea"
)

type Model struct{}

func (m Model) Init() tea.Cmd {
	return nil
}

func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
	return m, tea.Quit
}

func (m Model) View() string {
	return ""
}

func main() {
	_, err := tea.NewProgram(Model{}).Run()
	if err != nil {
		panic(err)
	}
}

```

Build the binary for linux: `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o ./main ./main.go`

Run in Docker: `docker run -v "$(pwd)"/main:/usr/local/bin/my-cli public.ecr.aws/ubuntu/ubuntu:22.04 my-cli`

```
panic: open /dev/tty: no such device or address

goroutine 1 [running]:
main.main()
	/Users/jonbretman/code/bubbletea-tty-issue/main.go:24 +0x53
```

I'm using AWS Ubuntu because I discovered this issue while using my CLI tool in AWS CodeBuild, but the issue is not specific to AWS or CodeBuild (as far as I can tell). Running with `-it` works e.g. `docker run -it ...`

**Expected behavior**
I would expect the program to still run

meowgorithm:
This came up [in Discord](https://discord.com/channels/1032368467246075914/1123965681549967440/1123965681549967440) recently as well. Based on this and that I don't think we should fatally error if we're not able to open a TTY. To help work around any resulting confusion we may want to consider:

1. Exposing input on `Program` (i.e. `Program.Input() io.Reader` … `nil` would mean no input)
2. Send something like a `WarningNoInputMsg` on program start if no input can be detected.

Also note that internally, if input's not a TTY, we try and open a TTY at `/dev/tty` (or `CONIN$` on Windows).

Fwiw, to some degree you can work around this in the meantime by detecting presence of a TTY prior to starting the program and passing `tea.WithInput(nil)` to `NewProgram` if no TTY is available.

DavidS-ovm:
[...]

PR #1680, opened 2026-04-19
fix: avoid Exec restore panic with WithInput(nil)

lawrence3699:

## Summary

Fix the `WithInput(nil)` + `ExecProcess` panic reported in the `#761` discussion.

## Problem

`#761` recommends `tea.WithInput(nil)` as the workaround when there is no TTY available. That works until a program executes `tea.ExecProcess(...)` and returns to Bubble Tea.

At that point `RestoreTerminal()` always called `initInputReader(false)`, even when input had been explicitly disabled. That rebuilds the cancel reader with `p.input == nil` and panics in `cancelreader`.

Minimal repro:

```go
    return tea.ExecProcess(exec.Command("true"), func(err error) tea.Msg {
        return finishedMsg{err}
    })
}

func main() {
    _, err := tea.NewProgram(model{}, tea.WithInput(nil)).Run()
    if err != nil {
        panic(err)
    }
}
```

## Fix

Make `RestoreTerminal()` match the initial startup path in `Run()` and only reinitialize the input reader when `p.input != nil`.

Added `TestTeaExecWithNilInput` to cover the regression.

## Validation

- `go test -run 'TestTeaExec|TestTeaExecWithNilInput' .`
- `go test ./...`

### DIFF
--- a/exec_test.go
+++ b/exec_test.go
@@ -14,9 +14,17 @@ type testExecModel struct {
 	err error
 }
 
+type testExecNoInputModel struct{ testExecModel }
+
 func (m *testExecModel) Init() Cmd {
 	c := exec.Command(m.cmd) //nolint:gosec
 	return ExecProcess(c, func(err error) Msg {
+		return execFinishedMsg{err}
+	})
+}
+
+func (m *testExecNoInputModel) Init() Cmd {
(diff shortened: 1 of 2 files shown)


aymanbagabas:
@lawrence3699 Thank you so much!

PR #1695, opened 2026-05-10
chore: remove deprecated windows note

Jaeiya:
I'm not sure if this is no longer an issue for all windows versions or just windows 10 & 11. I can only confirm 10 & 11. Let me know if there are any other considerations that need to be added or if we can just remove the note entirely, like I've done.

- [x] This pull request was approved to be created from a Discord conversation.

### DIFF
--- a/screen.go
+++ b/screen.go
@@ -3,9 +3,7 @@ package tea
 import "github.com/charmbracelet/x/ansi"
 
 // WindowSizeMsg is used to report the terminal size. It's sent to Update once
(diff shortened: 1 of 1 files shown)


meowgorithm (approved):

meowgorithm:
We added the ability for Windows to detect terminal size changes a very long time ago, and it should work for both Windows 10 and 11. Thanks for this!

andrinoff (approved):

PR #1703, opened 2026-05-27
docs: fix Init signature and Cmd usage in godoc examples

c-tonneslan:
Hit the BackgroundColorMsg example doc and noticed it wouldn't compile against v2:

```go
func (m Model) Init() (Model, Cmd) {
  return m, RequestBackgroundColor()
}
```

Two issues. The Model.Init signature is `Init() Cmd` now (single return, like all the examples in examples/), and `RequestBackgroundColor()` returns a Msg, not a Cmd, so it won't even type-check as the second return. `RequestBackgroundColor` (no parens) is itself the Cmd to hand back.

Same stale `(Model, Cmd)` Init signature showed up in three doc examples in commands.go (Batch, Every, Tick). Updated all four. While in there added the missing `return m, nil` on the BackgroundColorMsg Update example's fall-through path.

Sanity-checked by pasting the fixed BackgroundColorMsg snippet into a fresh `tea.Model` implementation against v2.0.6 and confirming `go build` is clean.

### DIFF
--- a/color.go
+++ b/color.go
@@ -54,8 +54,8 @@ func (e ForegroundColorMsg) IsDark() bool {
 // [BackgroundColorMsg.IsDark] to determine if the color is dark or light. For
 // example:
 //
-//	func (m Model) Init() (Model, Cmd) {
-//	  return m, RequestBackgroundColor()
+//	func (m Model) Init() Cmd {
+//	  return RequestBackgroundColor
 //	}
 //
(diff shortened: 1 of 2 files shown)


ossenthusiast:
I think the entire example should be removed since a lot of examples in examples directory show this exact same thing. Do instead of manually writing it here again: you could just link to it.

PR #1717, opened 2026-06-13
docs: fix Initialization heading level and de-duplicate Init prose

yanmxa:
The tutorial's Initialization section had two issues:

- The `Init` explanation ("`Init` can return a `Cmd` … return `nil`") appeared twice, and the first copy sat before `initialModel()`, which doesn't return a `Cmd`. Replaced it with a sentence about the initial state and kept the `Init` explanation once, where it belongs.
- `## Initialization` was an `h2`, which closed the `## Tutorial` section early and nested *The Update Method*, *The View Method*, and *All Together Now* under Initialization. Demoted it to `### Initialization` so those stay siblings under Tutorial.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,11 +96,10 @@ type model struct {
 }
 ```
 
-## Initialization
+### Initialization
 
-Next, we’ll define our application’s initial state. `Init` can return a `Cmd`
-that could perform some initial I/O. For now, we don’t need to do any I/O, so
-for the command, we’ll just return `nil`, which translates to “no command.”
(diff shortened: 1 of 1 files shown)


msdemers:
To add to @yanmxa PR, I wonder if this issue arrises from faulty copy/paste into the root README.md from the simplified version maintained at tutorials/basics/README.md. The description [here](https://github.com/charmbracelet/bubbletea/tree/main/tutorials/basics#initialization) looks correct to me.

PR #1721, opened 2026-06-16
fix: guard ProgressBarState.String against out-of-range values

### LINKED ISSUE #1711: ProgressBarState.String() panics on out-of-range values
**Describe the bug**

The method looks like this:

```go
    return [...]string{
        "None",
        "Default",
        "Error",
        "Indeterminate",
        "Warning",
    }[s]
}
```

---

**Setup**

- OS: not terminal-specific

---

**To Reproduce**

```go
package main

func main() {
    println(s.String())
}
```

2. Run `go run main.go`

3. Observe the output:

```
```

```
```

---

**Source Code**

```go
// tea.go, around line 321
    return [...]string{
        "None",
        "Default",
        "Error",
        "Indeterminate",
        "Warning",
    }[s]
}
```

---

**Expected behavior**

---

**Screenshots**

N/A

---

**Additional context**

N/A

truffle-dev:
`ProgressBar.State` is an exported field, so it can be assigned any `int` without going through `NewProgressBar`. `ProgressBarState.String()` indexed a fixed length-5 array literal directly with the receiver, which panics for any value outside `[0, 4]`:

```go
_ = s.String() // panic: runtime error: index out of range [5] with length 5
```

This bounds-checks the receiver and falls back to the stringer-style `ProgressBarState(N)` for unknown values, so `String()` never panics. Added a table-driven test covering the five defined states plus an above-range and a negative value.

`go test ./...` passes; `gofmt`/`go vet` clean.

Fixes #1711

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -317,8 +317,13 @@ const (
 	ProgressBarWarning
 )
 
-// String return a human-readable value for the given [ProgressBarState].
+// String returns a human-readable value for the given [ProgressBarState].
+// Values outside the defined range are rendered as ProgressBarState(N) rather
+// than panicking, since [ProgressBar.State] is exported and may hold any int.
 func (s ProgressBarState) String() string {
+	if s < ProgressBarNone || s > ProgressBarWarning {
+		return fmt.Sprintf("ProgressBarState(%d)", int(s))
+	}
 	return [...]string{
 		"None",
 		"Default",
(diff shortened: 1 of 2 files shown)


andrinoff:
Thank you for your contribution! 

Unfortunately, I am closing this PR in favor of #1748

PR #1744, opened 2026-07-25
examples/sequence: fix milisecond typo in SleepPrintln

makoto-developer:
The `SleepPrintln` helper in the `sequence` example spells its parameter
`milisecond` instead of `millisecond`. This renames it to the correct
spelling (both the declaration and its use inside the returned command).

Docs/example-only change; no behavior change.


meowgorithm:
My goodness. Thank you!

PR #1750, opened 2026-08-02
fix: restore kitty keyboard stack on exit

### LINKED ISSUE #1749: v2: kitty keyboard protocol is left active after exit
**Describe the bug**
Kitty keyboard protocol is left active after exit.

Keyboard enhancements are turned off with `^[[=0;1u` instead of popping the stack entry with `^[[<u`. This means that after exit, pressing keys like Home or Up sends `^[[1;1H^[[1;1A`, as shown by `cat -v`, instead of the normal `^[[H^[[A`. I can fix the terminal state by manually popping the state with `printf '\e[<u'`.

**Setup**
 - OS: CachyOS
 - Shell: zsh
 - Terminal Emulator: wezterm 20260716-195552-76b606ec with `enable_kitty_keyboard = true`

**To Reproduce**
1. Run any program that uses Bubble Tea v2 with keyboard enhancements.
2. Exit the program.
3. Press keys like Home or Up in the terminal.
4. The keys don't work as expected.

**Expected behavior**
Keys like Home or Up work normally in the terminal after a Bubble Tea v2 program exits.

**Additional context**
I encountered this bug when using Crush.

deepakganesh78:
Fixes #1749

## Reproduction
Issue #1749 reports that Bubble Tea v2 enables Kitty keyboard enhancements but exits with `CSI = 0 ; 1 u`, leaving terminals such as WezTerm in the enhanced keyboard state. The added `TestCursedRenderer_restoresKittyKeyboardStack` reproduces this programmatically by flushing and closing a renderer and asserting that Bubble Tea pushes a Kitty keyboard stack entry and pops it on close instead of resetting it in place.

## Root cause
The renderer used `ansi.KittyKeyboard(flags, 1)` to set the active keyboard flags and `ansi.KittyKeyboard(0, 1)` to clear them. That overwrites the terminal's current keyboard protocol flags instead of using the Kitty keyboard stack, so Bubble Tea cannot restore the prior terminal state on exit or when switching between main and alternate screen registries.

## Fix
Use `ansi.PushKittyKeyboard(flags)` when Bubble Tea takes ownership of a screen's keyboard enhancements, keep in-place updates for enhancement changes while the same screen is active, and use `ansi.PopKittyKeyboard(0)` when switching screens or closing. The first render directly into the alternate screen avoids popping the main-screen stack when Bubble Tea never pushed it. Renderer golden files were updated for the new escape sequences.

## Compatibility notes
This preserves Bubble Tea's existing default keyboard enhancement flags while making cleanup stack-balanced for terminals that implement the Kitty keyboard protocol. Non-supporting terminals should continue to ignore these sequences. `modifyOtherKeys` cleanup is unchanged.

## Validation
- `go test . -run TestCursedRenderer_restoresKittyKeyboardStack -count=1` passes with the fix.
- The regression test fails when `cursed_renderer.go` is reverted to the previous behavior.
- `gofmt -l .` produced no output.
- `go build ./...` passed.
- `go vet ./...` passed.
- `go test ./...` passed.


meowgorithm:
_Super good catch._

cursed_renderer.go
```diff
@@ -386,7 +386,11 @@ func (s *cursedRenderer) flush(closing bool) error {
 		_, _ = s.scr.WriteString(ansi.SetModifyOtherKeys2)
 
 		kittyFlags := keyboardEnhancementsFlags(view.KeyboardEnhancements)
-		_, _ = s.scr.WriteString(ansi.KittyKeyboard(kittyFlags, 1))
+		if s.lastView == nil || view.AltScreen != s.lastView.AltScreen {
+			_, _ = s.scr.WriteString(ansi.PushKittyKeyboard(kittyFlags))
+		} else {
+			_, _ = s.scr.WriteString(ansi.KittyKeyboard(kittyFlags, 1))
+		}
 		if !closing {
```
andrinoff, line 439: Two things here.

First: nothing tests this else branch. I thought `kitty_stop_startreleases` did, since it renders one view and then a second with `keyReleases: true`. But its golden only has `\x1b[>3u`. If the two views were flushed separately you'd see `\x1b[>1u` first, then `\x1b[=3;1u`. So they get merged into one flush and this branch never runs. That was already true before your PR, but it means the new code is uncovered.

Second: `KittyKeyboard(flags, 1)` changes whatever entry is on top of the stack right now. Usually that's ours. But if a nested `tea.Program` or a child process pushed an entry and never popped it, theirs is on top and we'd overwrite it. Pop-then-push can't do that. Skipping a push/pop on every keyboard change is a fair reason to do it this way -- could you add a comment saying that? Right now it looks like it contradicts the rest of the PR.

cursed_renderer.go
```diff
@@ -153,10 +153,10 @@ func (s *cursedRenderer) close() (err error) {
 		// two registries for the main and alt screens. We disable keyboard
 		// enhancements whenever we enter/exit alt screen mode in
 		// [cursedRenderer.flush].
-		// Here, we reset the keyboard protocol of the last screen used
-		// assuming the other screen is already reset when we switched screens.
+		// Here, we pop the keyboard protocol of the last screen used
+		// assuming the other screen is already popped when we switched screens.
 		_, _ = s.buf.WriteString(ansi.ResetModifyOtherKeys)
-		_, _ = s.buf.WriteString(ansi.KittyKeyboard(0, 1))
+		_, _ = s.buf.WriteString(ansi.PopKittyKeyboard(0))
```
andrinoff: Small thing: `PopKittyKeyboard(0)` looks like "pop zero entries." It actually sends `CSI < u`, and the spec says a missing number means 1, so it works. But `PopKittyKeyboard(1)` says what you mean and doesn't rely on the terminal getting the default right. Same at 523, and at 107 in the test.

cursed_renderer.go
```diff
@@ -511,11 +515,13 @@ func (s *cursedRenderer) flush(closing bool) error {
...
-		// We always disable keyboard enhancements when switching screens
-		// because the terminal is expected to have two different keyboard
-		// registries for main and alt screens.
-		_, _ = buf.WriteString(ansi.ResetModifyOtherKeys)
-		_, _ = buf.WriteString(ansi.KittyKeyboard(0, 1))
+		if s.lastView != nil {
+			// We always disable keyboard enhancements when switching screens
+			// because the terminal is expected to have two different keyboard
+			// registries for main and alt screens.
+			_, _ = buf.WriteString(ansi.ResetModifyOtherKeys)
```
andrinoff: `ResetModifyOtherKeys` is a different protocol from kitty. It has no stack, and sending `CSI > 4 m` to a fresh terminal does nothing. So it doesn't need the `s.lastView != nil` guard — only the pop below does. This is also what changes the first few bytes of `altscreen_autoexit.golden`. Any reason not to leave this line outside the if?

cursed_renderer.go
```diff
@@ -136,7 +136,7 @@ func (s *cursedRenderer) start() {
 	_, _ = s.scr.WriteString(ansi.SetModifyOtherKeys2)
 
 	kittyFlags := keyboardEnhancementsFlags(s.lastView.KeyboardEnhancements)
-	_, _ = s.scr.WriteString(ansi.KittyKeyboard(kittyFlags, 1))
+	_, _ = s.scr.WriteString(ansi.PushKittyKeyboard(kittyFlags))
```
andrinoff: A one-line comment would help a lot next time someone touches this.

cursed_renderer_test.go
```diff
@@ -78,3 +82,33 @@ func TestCursedRenderer_mouseVsFlush(t *testing.T) {
...
+
+func TestCursedRenderer_restoresKittyKeyboardStack(t *testing.T) {
+	t.Parallel()
+
+	var out bytes.Buffer
+	r := newCursedRenderer(&out, []string{"TERM=xterm-256color"}, 80, 24)
+
+	view := NewView("hello")
+	view.KeyboardEnhancements.ReportEventTypes = true
+	r.render(view)
```
andrinoff: This test never calls `r.start()`, so the push inside `start()` isn't covered. That's the one that runs on resume and after `ExecProcess`. It also never turns alt screen on or off, which is the hardest part of this change.

cursed_renderer_test.go
```diff
@@ -78,3 +82,33 @@ func TestCursedRenderer_mouseVsFlush(t *testing.T) {
...
+
+	got := out.String()
+	wantPush := ansi.PushKittyKeyboard(keyboardEnhancementsFlags(view.KeyboardEnhancements))
+	if !strings.Contains(got, wantPush) {
+		t.Fatalf("expected kitty keyboard protocol to be pushed with %q, got %q", wantPush, got)
+	}
+	wantPop := ansi.PopKittyKeyboard(0)
+	if !strings.Contains(got, wantPop) {
+		t.Fatalf("expected kitty keyboard protocol to be popped with %q, got %q", wantPop, got)
+	}
```
andrinoff: Both checks are `strings.Contains`, so the test passes even if the pop comes before the push. Something like `strings.Index(got, wantPush) < strings.Index(got, wantPop)` would catch that. Wrong order is the bug most likely to show up here later.

andrinoff (changes requested):
Hey there @deepakganesh78! 

Thank you for your contribution! You've beat me to it 😆 

Please, note the review comments below, and try to fix it.

p.s. great work!

andrinoff (approved):

andreynering (approved):

PR #1751, opened 2026-08-03
fix(kitty-kb-protocol): properly pop off the stack

### LINKED ISSUE #1749: v2: kitty keyboard protocol is left active after exit

andrinoff:
## What?

Replace the Kitty keyboard protocol teardown sequence `CSI = 0 ; 1 u` (set flags to zero)
with a proper stack pop `CSI < u`, and switch the enable path to use `CSI > flags u`
(push) so each enable is paired with exactly one pop. Golden test files regenerated to
match.

## Why?

Closes #1749


andrinoff:
(@)deepakganesh78 beat me to it :X

PR #1757, opened 2026-08-08
fix(key): map media record to ultraviolet code

Ricardo-M-L:
## Problem

`KeyMediaRecord` omits its initializer in the media-key constant block. In a Go const declaration, that repeats the previous expression rather than incrementing it, so Bubble Tea currently exports:

```go
tea.KeyMediaRecord == tea.KeyMediaPrev
```

Ultraviolet decodes the record key as `uv.KeyMediaRecord`, which means Bubble Tea applications cannot match that event using the advertised `tea.KeyMediaRecord` constant.

This is the same Go const trap recently identified for `MouseButton11` in #1754, but in the independent media-key block.

## Fix

Explicitly map `KeyMediaRecord` to `uv.KeyMediaRecord`, matching every other re-export in the block.

The regression test checks the complete media-key block against Ultraviolet so another omitted or mistyped mapping is caught directly.

## Validation

- `go test ./...`
- `go vet ./...`
- `git diff --check`
- Fork CI on this exact head commit:
  - [build](https://github.com/Ricardo-M-L/bubbletea/actions/runs/31248273577) — Linux/macOS/Windows root and examples builds/tests plus `govulncheck`
  - [lint](https://github.com/Ricardo-M-L/bubbletea/actions/runs/31248273602) — Linux/macOS/Windows
  - [coverage](https://github.com/Ricardo-M-L/bubbletea/actions/runs/31248273453) — race-enabled coverage


andrinoff:
Thank you for your contribution, @Ricardo-M-L 

Please, remove the test.

Ricardo-M-L:
Done — removed the test in the latest push. Thanks for the review!

andrinoff (approved):

PR #1758, opened 2026-08-08
fix: avoid blocking Program.Print methods after exit

Ricardo-M-L:
## Problem

`Program.Println` and `Program.Printf` send directly to the unbuffered `p.msgs` channel. After `Run` returns, the event loop is gone and there is no receiver, so either method blocks forever when called during late goroutine cleanup or after program shutdown.

This differs from `Program.Send`, which already documents and implements post-exit calls as a no-op by selecting on the program context.

## Fix

Route both print methods through `Program.Send` and document the post-exit behavior. Printing while the program is running is unchanged; after exit the methods now return instead of leaking or deadlocking a goroutine.

The regression test runs a program to completion and verifies both methods return. Against `main`, each call remains blocked on `p.msgs`.

## Validation

- `go test -run TestProgramPrintAfterExit -count=100 ./...`
- `go test -race -run TestProgramPrintAfterExit -count=50 ./...`
- `go test ./...`
- `go vet ./...`
- `git diff --check`

### DIFF
--- a/tea.go
+++ b/tea.go
@@ -1369,10 +1369,11 @@ func (p *Program) RestoreTerminal() error {
 // and will persist across renders by the Program.
 //
 // If the altscreen is active no output will be printed.
+// If the program has already exited this is a no-op.
 func (p *Program) Println(args ...any) {
-	p.msgs <- printLineMessage{
+	p.Send(printLineMessage{
 		messageBody: fmt.Sprint(args...),
-	}
+	})
 }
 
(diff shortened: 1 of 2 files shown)


Ricardo-M-L:
Friendly ping — this is a small, isolated fix with no behavioral changes beyond preventing deadlocks on `Program.Print` after exit. Happy to address any concerns or split if needed. Would love to see this land in the next release!

PR #1770, opened 2026-08-17
fix: properly clean up renderer and reader on panic

Ricardo-M-L:
This PR addresses: properly clean up renderer and reader on panic


Ricardo-M-L:
Friendly ping - this PR is a single-hunk fix and ready for review.

**What it fixes:** `recoverFromGoPanic` only called `p.cancel()`, which stops the message loop but leaves the renderer goroutine and the input reader open. On a panic that leaked a goroutine and held the reader until process exit. Switching to `p.shutdown(true)` runs the full teardown path (renderer, input reader, terminal state) that the normal exit already uses.

**Scope:** 1 hunk in `tea.go`. The `tutorials/go.mod` / `go.sum` changes are from a `go mod tidy` while building the examples - drop them on request to keep this a single-file PR.

**Overlap note:** the same fix is also present on the branch for PR 1779 (which bundles this with the Println/Printf deadlock fix). If you'd rather this land on its own, I can drop the hunk from 1779 - just say which of the two you'd prefer to keep open.

PR #1776, opened 2026-08-20
perf: render only when the view changes

anandh8x:
## What changed

- Wake the renderer when a new view is queued instead of polling it continuously.
- Coalesce bursts and retain `WithFPS` as the maximum active render rate.
- Wake the renderer after terminal restoration so resumed programs repaint normally.
- Add regression coverage for idle sleep, view-change wakeups, burst coalescing, and queued terminal output.

## Why

The current renderer starts a fixed-rate ticker for the full lifetime of every program. Even when the model schedules no commands and its view never changes, the process wakes and compares the frame at the configured FPS.

This keeps the same initial-frame ordering and active frame-rate ceiling, but lets the renderer block after the screen becomes unchanged. The next view change wakes it immediately; updates arriving inside one frame interval are rendered together.

## Reproduction and result

Measured with a static Bubble Tea v2.0.9 model whose `View` never changes and whose `Init` and `Update` schedule no commands. Both binaries ran in the same 80x24 real PTY, settled for 3 seconds, and were sampled for 10 seconds.

| Renderer | Idle CPU | Startup |
| --- | ---: | ---: |
| v2.0.9 fixed ticker | 0.790% | 50 ms |
| Demand-driven scheduler | 0.000% | 41 ms |

In an animation-enabled interactive run, both versions emitted 139 active output events. Mean active gaps were 15.75 ms before and 15.68 ms after, so the change did not reduce active rendering cadence.

## Validation

- `go test ./... -count=1`
- `go vet ./...`
- `go test -race ./... -count=1`
- Repeated focused scheduler and renderer race tests
- Real-PTY idle and active rendering checks

### DIFF
--- a/renderer_scheduler_test.go
+++ b/renderer_scheduler_test.go
@@ -101,6 +101,10 @@ func TestRendererWakesForQueuedTerminalOutput(t *testing.T) {
 	t.Cleanup(func() { program.stopRenderer(true) })
 
 	program.execute("terminal query")
-	program.render(program.initialModel)
 	waitForSchedulerFlush(t, renderer)
+	program.mu.Lock()
+	defer program.mu.Unlock()
+	if program.outputBuf.Len() != 0 {
+		t.Fatalf("queued terminal output was not flushed: %q", program.outputBuf.String())
+	}
 }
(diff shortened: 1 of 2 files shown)


anandh8x:
Fixed one missed wake-up case: queued terminal sequences now wake the renderer, so they flush even when no view render follows. Added a regression test for it.

PR #1779, opened 2026-08-25
fix: prevent deadlock in Println/Printf after program shutdown

Ricardo-M-L:
Fix Program.Println() and Printf() which sent directly to p.msgs without context check, unlike Send(). After shutdown these calls blocked forever, violating the documented safety contract.


Ricardo-M-L:
Friendly ping - this PR is small and ready for review whenever there's bandwidth.

**What it fixes:** `Program.Println` / `Printf` send via `p.msgs <-` directly, which is an unbuffered channel send. Once the program has exited, the `eventLoop` goroutine that drains `p.msgs` is gone, so a late `Println` blocks forever and leaks the calling goroutine (deadlock). The fix routes through `p.Send(...)`, which is already guarded against a closed/stopped program.

**Scope:** 1 hunk in `tea.go`. The `tutorials/go.mod` / `go.sum` touchups are from a `go mod tidy` while building the examples - happy to drop them if you'd rather keep this PR to a single file.

**Overlap note:** this branch also carries the `recoverFromGoPanic` cleanup fix (`p.cancel()` -> `p.shutdown(true)`) which is filed separately as PR 1770. If you'd prefer one PR per fix, I can split that hunk out so this one is only the Println/Printf change.

PR #1798, opened 2026-09-09
fix(renderer): don't trim final newline (because it can be intentional)

meowgorithm:
This revision maintains trailing newlines, because trailing blank lines can be a design choice. Prior to this change it was nearly impossible to keep them.

Worth testing this one thoroughly, per the removed comment in the code.


cursed_renderer.go
```diff
@@ -514,16 +514,28 @@ func (s *cursedRenderer) flush(closing bool) error {
...
+		// terminals.
+		//
+		// The frame's height counts the view's trailing blank lines, so
+		// parking here also commits the renderer to them: a frame ending
+		// in intentional blank lines keeps its bottom margin, and output
+		// printed after the program exits starts below the frame rather
+		// than on its last visible line. Frames taller than the screen
+		// have their top rows dropped, so their last visible row is the
+		// screen's.
+		y := frameArea.Dy() - 1
```
andrinoff, line 531: parking at the frame's last row is correct, but for frames taller than the screen this now aligns the frame with the top of the terminal, and the existing per-frame full redraw then floods scrollback in tmux

cursed_renderer_test.go
```diff
@@ -95,6 +95,79 @@ func assertInOrder(t *testing.T, got string, wants ...string) {
...
+		t.Fatalf("cursor parked at (%d, %d), want (0, 0)", x, y)
+	}
+	if err := r.close(); err != nil {
+		t.Fatal(err)
+	}
+}
+
+// Frames taller than the screen drop their top rows, so the parked cursor
+// must land on the screen's last visible row, not beyond it.
+func TestCursedRenderer_trailingBlankLinesTallFrame(t *testing.T) {
```
andrinoff, line 148: this checks the parked position but not that a second unchanged flush is a no-op. that's why the regression above was missed

andrinoff (changes requested):
There's also a pre-existing bug that a view without a trailing newline loses its last line on exit. `close` moves to the last buffer row and erases below it, so the shell prompt overwrites that line. Maybe worth fixing here too

PR #1801, opened 2026-09-11
fix: skip capability queries when input is disabled

drawmeanelephant:
## What

Skip the synchronized-output (mode 2026) and unicode-core (mode 2027) capability query when input is disabled (`WithInput(nil)`), mirroring what #1750 did for the kitty keyboard query.

## Why

Programs without input can never read the terminal's reply, so the response leaks into the shell once the program exits. For short-lived programs — e.g. `gum spin` wrapping a fast command — the query can even be emitted during shutdown, after the renderer has restored the terminal, so the leak is deterministic rather than a timing race:

```text
$ script -q /dev/null gum spin --spinner dot --title x -- bash -c 'echo hi'
...^[[?25h^[[?2004l^[[?2026$p^[[?2027$p^[[?2026;2$y^[[?2027;1$y
```

Since the reply can never be consumed, the renderer can never enable the modes anyway, so skipping the query loses nothing.

Refs charmbracelet/gum#1118, charmbracelet/gum#1138, #1590.

## Validation

- Added `TestTeaNoCapabilityQueryWithoutInput`; it fails on the input-disabled case without the guard and passes with it.
- `gofmt -l .`, `go vet ./...`, `go test ./...` all pass.


andrinoff (changes requested):
Other terminal questions (background color, clipboard, cursor position, around lines 819–864) can still be sent with input disabled and would leak the same way.

andrinoff:
Also, please, rebase onto the current main for CI

drawmeanelephant:
Addressed both in b6862bd: all input-dependent queries (clipboard reads, background/foreground/cursor colors, cursor position, terminal version, termcap, and the startup mode 2026/2027 query) now route through a shared `executeQuery` that skips them when input is disabled, with a regression test covering each one. Also rebased onto current `main` (657aaae).

Heads-up: the workflow runs for the new commit are sitting in `action_required` and need a maintainer to approve them before CI starts. Thanks for the review!

hanoii:
Reproduces with gum v2.0.1 (bubbletea v2.0.9) in kitty, inside a DDEV container. Every `gum spin` leaves `^[[?2026;2$y^[[?2027;0$y` on the shell prompt when it finishes, since spin uses `tea.WithInput(nil)` and nothing reads the reply. Skipping the query when input is nil is the right fix for this case. Looking forward to it landing.

PR #1813, opened 2026-09-20
fix: stop Sequence after Quit, Interrupt, or shutdown

kevin9327:
## Summary

`tea.Sequence` ran every remaining command even after the program was already quitting.

That happens because `execSequenceMsg` treated `QuitMsg` / `InterruptMsg` like any other message: it sent them and then kept executing later commands. The same thing happened after shutdown cancelled the program context. Nested sequences had the same leak: `Sequence(Sequence(Quit), later)` still ran `later`.

Those later commands still performed their side effects (I/O, timers, subprocesses) after the UI had begun to exit.

## Fix

- Return from `execSequenceMsg` after sending `QuitMsg` or `InterruptMsg`.
- Return immediately when the program context is already done, so commands queued behind a cancelled sequence do not run.
- Propagate that stop through nested `sequenceMsg` values.

## Test plan

- [x] `TestTeaSequenceStopsAfterQuit`
- [x] `TestTeaNestedSequenceStopsAfterQuit`
- [x] `TestTeaSequenceStopsOnContextDone`
- [x] Existing sequence/batch tests
- [x] `go vet` / `gofmt`

AI-assisted (Grok)


otrumb (changes requested):
Blocker: `tea.go` terminal-message branch stops sequence after `p.Send` even when `WithFilter` rejects or replaces `QuitMsg`.

Evidence:
- Real `Program.Run` reproduction using `Sequence(Quit, later)` with a filter returning `nil` failed 20/20.
- Removing only the new terminal branch passed 20/20.
- Existing suite, repeated focused tests, and `go vet`/build with `CGO_ENABLED=0` pass.

Please preserve sequence continuation when a terminal message is rejected or replaced by `WithFilter`, and add a regression test. Reviewed head: `838bd6cf5a5afe28fbf45e689dd2222e228f88e9`.

